PR #26, opened 2026-04-19
Add core library and app framework foundation

rjwalters:
## Summary

- Adds the core `sokol_gfx` equivalent: typed resource handles, buffer/image/sampler/shader/pipeline creation, render pass management, uniform binding with dynamic offsets, and draw call dispatch
- Adds the `sokol_app` equivalent: async WebGPU initialization, requestAnimationFrame loop, input events (keyboard, mouse, touch, resize), high-DPI handling, and cleanup
- Defines the full type system: enums for vertex formats, pixel formats, blend/depth/cull modes, and descriptor interfaces for all resources

## Files

| File | Purpose |
|------|---------|
| `src/types.ts` | All resource handles, enums, descriptor interfaces, `Gfx` and `AppEvent` types |
| `src/gfx.ts` | `createGfx()` — WebGPU resource management and rendering implementation |
| `src/app.ts` | `run()` — app lifecycle, WebGPU init, input events, frame loop |
| `src/index.ts` | Public API re-exports |
| `tsconfig.json` | TypeScript config (strict, ESNext, bundler resolution) |

## Test plan

- [ ] `npx tsc --noEmit` passes with no type errors
- [ ] Review all WebGPU API usage for correctness
- [ ] Verify resource handle branding prevents cross-type misuse
- [ ] Confirm descriptor defaults match Sokol conventions

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,0 +1,103 @@
+import { type AppDesc, type AppEvent, AppEventType } from "./types.js";
+import { createGfx } from "./gfx.js";
+
+export async function run(desc: AppDesc): Promise<() => void> {
+  const canvas = typeof desc.canvas === "string"
+    ? document.querySelector<HTMLCanvasElement>(desc.canvas)!
+    : desc.canvas;
+
+  if (!canvas) throw new Error("Canvas not found");
+
+  if (!navigator.gpu) {
+    throw new Error("WebGPU is not supported in this browser");
+  }
+
+  const adapter = await navigator.gpu.requestAdapter();
+  if (!adapter) throw new Error("Failed to get GPU adapter");
+
+  const device = await adapter.requestDevice();
+  const context = canvas.getContext("webgpu")!;
+  const format = navigator.gpu.getPreferredCanvasFormat();
+
+  context.configure({ device, format, alphaMode: "premultiplied" });
+
+  const pixelRatio = desc.pixelRatio ?? window.devicePixelRatio;
+
+  function resize() {
+    const w = Math.floor(canvas.clientWidth * pixelRatio);
+    const h = Math.floor(canvas.clientHeight * pixelRatio);
+    if (canvas.width !== w || canvas.height !== h) {
+      canvas.width = w;
+      canvas.height = h;
+    }
+  }
+
+  resize();
+
+  const gfx = createGfx(device, canvas, context, format);
+
+  await desc.init(gfx);
+
+  // Input events
+  function dispatch(ev: AppEvent) {
+    desc.event?.(ev, gfx);
+  }
+
+  const eventMap: [EventTarget, string, (e: Event) => void][] = [];
+
+  function listen<K extends keyof HTMLElementEventMap>(
+    target: EventTarget,
+    type: K,
+    handler: (e: HTMLElementEventMap[K]) => void,
+  ) {
+    const h = handler as (e: Event) => void;
+    target.addEventListener(type, h);
+    eventMap.push([target, type, h]);
+  }
+
+  listen(canvas, "mousedown", (e) => dispatch({ type: AppEventType.MOUSE_DOWN, mouseX: e.offsetX * pixelRatio, mouseY: e.offsetY * pixelRatio, mouseButton: e.button }));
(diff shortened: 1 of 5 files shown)


rjwalters:
## Loom Review: Add core library and app framework foundation

**Verdict: Approve** -- This is a solid foundation with good Sokol-style API design. The issues below are worth tracking but should not block the merge since many are explicitly scoped for follow-up issues.

### What works well

- **Branded handles** -- The `_brand` pattern for type-safe resource handles is well done and prevents cross-type misuse at compile time.
- **Sokol-faithful API surface** -- `beginPass` / `applyPipeline` / `applyBindings` / `applyUniforms` / `draw` / `endPass` / `commit` mirrors Sokol's render flow accurately.
- **Clean separation** -- `types.ts` (interfaces/enums), `gfx.ts` (GPU resource management), `app.ts` (lifecycle/events) is a natural split.
- **Proper WebGPU init** -- Adapter/device negotiation, canvas context configuration, ResizeObserver, and high-DPI handling are all correct.
- **Dynamic uniform offsets** -- The 256-byte alignment for dynamic uniform offsets correctly follows the WebGPU spec's `minUniformBufferOffsetAlignment`.

### Issues to address (future PRs are fine)

**Bug: `@webgpu/types` conflicts with built-in DOM types (High)**

`tsconfig.json` specifies `"types": ["@webgpu/types"]`, but modern TypeScript (5.x+) already ships WebGPU types in `lib.dom.d.ts`. This causes `tsc --noEmit` to fail with dozens of duplicate declaration errors. Either:
- Remove `@webgpu/types` from devDependencies and the `types` array, or
- Add `"skipLibCheck": true` to tsconfig (less ideal)

**Bug: Non-assertive null check on canvas (Low)**

```typescript
const canvas = typeof desc.canvas === "string"
  ? document.querySelector<HTMLCanvasElement>(desc.canvas)!  // non-null assertion
  : desc.canvas;
if (!canvas) throw new Error("Canvas not found");
```
The `!` on line 6 of `app.ts` defeats the null check on line 9. If the selector doesn't match, TypeScript won't warn because of the assertion. Remove the `!` and let the null check do its job.

**Design: `gpuBufferUsage` is a no-op switch (Low)**

```typescript
function gpuBufferUsage(usage: BufferUsage | undefined): number {
  const base = GPUBufferUsage.COPY_DST;
  switch (usage) {
    default: return base | GPUBufferUsage.VERTEX | GPUBufferUsage.INDEX;
  }
}
```
This always returns the same flags regardless of `BufferUsage`. The `IMMUTABLE` / `DYNAMIC` / `STREAM` enum values are defined but never differentiated. Fine as a stub if tracked for follow-up.

**Design: `Bindings.images` and `Bindings.samplers` are accepted but never used (Medium)**

`applyBindings()` handles vertex/index buffers but ignores `images` and `samplers`. Shaders that need textures will fail silently. The bind group layout in `makePipeline` also only creates group 0 (uniforms) with no group 1 for textures. This needs to be tracked.

**Design: `depthAttachment` in `PassDesc` is accepted but never used (Medium)**

`beginPass()` reads `colorAttachments` and `offscreen` but completely ignores `depthAttachment`. Any pipeline with depth testing enabled will fail at render time because no depth/stencil attachment is provided to the render pass.

**Design: Uniform buffer overflow not checked (Low)**

`applyUniforms()` writes into a fixed 64KB buffer without checking if `uniformOffset` exceeds `UNIFORM_BUFFER_SIZE`. Heavy scenes could silently corrupt data or crash.

**Design: `makeImage` hardcodes `bytesPerRow` to `width * 4` (Low)**

This only works for RGBA8/BGRA8 formats. For R8, RG8, RGBA16F, RGBA32F, the bytes-per-pixel differs. Fine for now, but worth noting.

**Nit: Touch event coordinates use `clientX/clientY` instead of canvas-relative offsets**

Mouse events correctly use `e.offsetX/offsetY` (canvas-relative), but touch events use `t.clientX/clientY` (viewport-relative). This will produce incorrect coordinates if the canvas is not at (0,0) in the page.

**Nit: `wheel` event listener casts to `WheelEvent` unnecessarily**

The `listen` call uses `"wheel"` as the event type, so the handler parameter `e` is already typed as `WheelEvent` via `HTMLElementEventMap["wheel"]`. The `as WheelEvent` casts are redundant.

### Summary

The foundation is well-structured and demonstrates a clear understanding of both Sokol's API philosophy and WebGPU's resource model. The `@webgpu/types` conflict should be fixed before merge (one-line tsconfig change). Everything else is either a stub for future work or a minor nit. Good to merge once the type-checking issue is resolved.

PR #27, opened 2026-04-19
Fix buffer usage flags and add partial update support

### LINKED ISSUE #2: Buffer creation, update, and destroy (vertex, index, uniform)
## Summary
Create, update, and destroy GPU buffers for vertex data, index data, and uniforms.

## Current State
`gfx.ts` implements `makeBuffer`, `updateBuffer`, `destroyBuffer`. Supports initial data upload via `mappedAtCreation`.

## Requirements
- [x] Create buffers with initial data (`gfx.ts:92-107`)
- [x] Update buffer contents via `queue.writeBuffer` (`gfx.ts:243-247`)
- [x] Destroy buffers and release pool slots (`gfx.ts:231-234`)
- [ ] Proper usage flag mapping (`gpuBufferUsage` ignores the `usage` parameter — `gfx.ts:77-81`)
- [ ] Buffer size validation and alignment
- [ ] Partial buffer updates (offset + size)
- [ ] Buffer usage tracking to prevent invalid operations

## References
- Sokol: `sg_make_buffer()`, `sg_update_buffer()`, `sg_destroy_buffer()`
- WebGPU: `GPUDevice.createBuffer()`, `GPUQueue.writeBuffer()`

rjwalters:
Closes #2

## Changes

- **`gpuBufferUsage()` fix** (`src/gfx.ts`): Properly maps `BufferUsage.IMMUTABLE`/`DYNAMIC`/`STREAM` to `GPUBufferUsage` flags. `IMMUTABLE` buffers no longer include `COPY_DST` — data is supplied only at creation via `mappedAtCreation`.
- **`updateBuffer()` guard** (`src/gfx.ts`): Throws `"Cannot update an IMMUTABLE buffer"` when called on a buffer created with `BufferUsage.IMMUTABLE`.
- **Buffer size validation** (`src/gfx.ts`): `makeBuffer` now enforces a minimum of 4 bytes and rounds up to 4-byte alignment. `updateBuffer` also rounds the write size up to a multiple of 4 to satisfy WebGPU's `writeBuffer` constraint.
- **Partial update support** (`src/gfx.ts`, `src/types.ts`): `updateBuffer` accepts an optional `dstOffset` parameter (default `0`) passed directly to `GPUQueue.writeBuffer` as `bufferOffset`.


rjwalters:
## Loom Review — Changes Requested

**TypeScript:** clean (`tsc --noEmit` passes with zero errors).

The four stated goals are correct in spirit, but one implementation detail introduces a real correctness bug.

---

### Bug: `updateBuffer` over-reads source `ArrayBuffer` when size is not 4-byte aligned

**File:** `src/gfx.ts`, `updateBuffer`

```ts
const writeBytes = Math.ceil(data.byteLength / 4) * 4;
device.queue.writeBuffer(slot.gpu, dstOffset, data.buffer, data.byteOffset, writeBytes);
```

`GPUQueue.writeBuffer` validates that `data.byteOffset + writeBytes <= data.buffer.byteLength`. When the caller passes a standalone `ArrayBufferView` (the overwhelmingly common case) whose `byteLength` is not already a multiple of 4, rounding `writeBytes` up causes it to exceed the backing `ArrayBuffer` size. WebGPU will raise a validation error and the write will be silently dropped.

Example: `new Uint8Array(5)` — `byteLength = 5`, `writeBytes = 8`, but `buffer.byteLength = 5`. The call will fail.

**Fix:** Copy the data into a new, padded `ArrayBuffer` before passing to `writeBuffer`:

```ts
const writeBytes = Math.ceil(data.byteLength / 4) * 4;
const src = new Uint8Array(writeBytes);          // zero-padded
src.set(new Uint8Array(data.buffer, data.byteOffset, data.byteLength));
device.queue.writeBuffer(slot.gpu, dstOffset, src.buffer, 0, writeBytes);
```

This keeps the buffer-side alignment requirement satisfied while reading only bytes that actually exist in the source.

---

### Other observations (non-blocking)

- **`gpuBufferUsage` switch (`IMMUTABLE`):** Correct — omitting `COPY_DST` for IMMUTABLE buffers and relying on `mappedAtCreation` is exactly right per the WebGPU spec.
- **`makeBuffer` size alignment (`Math.max(4, Math.ceil(rawSize / 4) * 4)`):** Correct. Note that the stored `desc` still holds the original `rawSize` (or none) rather than `alignedSize`; the draw-count auto-derivation in `draw()` (on other branches) reads `desc.size`, so there may be an off-by-a-few-bytes edge case there if those features are ever rebased on top of this. Worth noting for the future, not a blocker here.
- **`dstOffset` alignment:** `writeBuffer` requires `bufferOffset` to be a multiple of 4. This is not enforced at the API boundary; callers who pass an unaligned offset will get a GPU validation error. A runtime `if (dstOffset % 4 !== 0) throw new Error(...)` guard would be a nice addition but is not required to unblock.
- **IMMUTABLE guard uses `=== BufferUsage.IMMUTABLE` (`=== 0`):** `undefined` usage falls through to the `default` case in `gpuBufferUsage` (gets `COPY_DST`) and is not treated as IMMUTABLE by the guard — which is the right behavior. Correct.

rjwalters:
## Loom Review — Changes Requested

**TypeScript:** `tsc --noEmit` passes cleanly (zero errors).

### Bug still present: `updateBuffer` over-reads source ArrayBuffer

The previously reported bug has **not been fixed**. The code in `updateBuffer` still reads:

```ts
const writeBytes = Math.ceil(data.byteLength / 4) * 4;
device.queue.writeBuffer(slot.gpu, dstOffset, data.buffer, data.byteOffset, writeBytes);
```

When `data.byteLength` is not a multiple of 4, `writeBytes` exceeds the available bytes in `data.buffer` starting from `data.byteOffset`. For example, `new Uint8Array(5)` has `buffer.byteLength = 5`, but `writeBytes = 8` — `writeBuffer` will throw a validation error because `byteOffset(0) + size(8) > buffer.byteLength(5)`.

**Required fix** (as described in the previous review): create a zero-padded copy when alignment padding is needed:

```ts
const writeBytes = Math.ceil(data.byteLength / 4) * 4;
if (writeBytes > data.byteLength) {
  const padded = new Uint8Array(writeBytes);
  padded.set(new Uint8Array(data.buffer, data.byteOffset, data.byteLength));
  device.queue.writeBuffer(slot.gpu, dstOffset, padded.buffer, 0, writeBytes);
} else {
  device.queue.writeBuffer(slot.gpu, dstOffset, data.buffer, data.byteOffset, writeBytes);
}
```

The fast path (`writeBytes === data.byteLength`) avoids the allocation when the data is already aligned.

Everything else in the PR looks correct (buffer usage flags, IMMUTABLE guard, size alignment in `makeBuffer`, `dstOffset` parameter).

rjwalters:
## Loom Judge Review

### Verdict: Changes Requested

### The targeted fixes are correct

The two specific changes this PR set out to make are sound:

1. **Buffer usage flags** -- `IMMUTABLE` correctly omits `COPY_DST`; `DYNAMIC`/`STREAM` correctly include it. The `updateBuffer` guard against immutable buffers is good.

2. **Padding fix** -- The zero-padding path is correct. When `writeBytes > data.byteLength`, creating a new `Uint8Array(writeBytes)` (zero-initialized by spec), copying the source data in, and passing the padded buffer to `writeBuffer` avoids the over-read. The fast path that skips padding when already aligned is a nice optimization.

3. **Buffer size alignment** -- `Math.max(4, Math.ceil(rawSize / 4) * 4)` correctly enforces WebGPU's minimum and alignment requirements.

4. **Partial update support** -- `dstOffset` parameter is cleanly added.

### Critical issue: branch is stale and would regress main

The branch diverged from an early version of `main` and is **missing 8 merged commits**. Merging this PR as-is would **revert** the following features already on `main`:

- `BufferUsage.INDIRECT` support and `drawIndirect()` (PR #36)
- Async `makeShader()` with compilation error checking (PR #32)
- Configurable shader entry points (`vertexEntry`/`fragmentEntry`)
- Draw call auto-derive, buffer capacity validation, and `DrawStats` (PR #36)
- Depth/stencil attachments, resolve targets, store actions, multi-pass encoder reuse (PR #42)
- `cssWidth`/`cssHeight`/`dpiScale` getters (PR #35)
- `sampleCount` on textures and `multisample` on pipelines
- `uniformOffset` reset in `commit()`

### Minor issue: `dstOffset` alignment not validated

WebGPU requires `bufferOffset` (the `dstOffset` parameter) to be a multiple of 4. The current code passes it directly to `writeBuffer` without validation. Consider adding: `if (dstOffset % 4 !== 0) throw new Error("dstOffset must be a multiple of 4");`

### Required action

Rebase this branch onto current `main` and re-apply the buffer usage, padding, and partial update changes to the current codebase. The fixes themselves are good -- they just need to land on top of the current code, not alongside regressions.

rjwalters (commented):
## Loom Judge Review: Fix buffer usage flags and add partial update support

**Verdict: APPROVED (ready to merge)**

### Scope

The PR consists of 2 commits modifying `src/gfx.ts` and `src/types.ts`:
1. `1965c57` — Fix buffer usage flags, add size validation and partial update support
2. `d4e8073` — Fix updateBuffer over-read when data size is not 4-byte aligned

(The merge commit `31d69b5` pulls in already-merged main changes and introduces no new conflicts.)

### Correctness

**gpuBufferUsage() fix** -- Correct. IMMUTABLE buffers correctly omit `COPY_DST` since data is provided only via `mappedAtCreation`. DYNAMIC/STREAM/default properly include `COPY_DST`. The INDIRECT case (added proactively before the enum value existed on this branch, now present after merge) correctly maps to `COPY_DST | INDIRECT`.

**IMMUTABLE guard in updateBuffer()** -- Correct. Prevents a silent WebGPU validation error with a clear thrown message. The guard `slot.desc.usage === BufferUsage.IMMUTABLE` works correctly even when `usage` is `undefined` (since `undefined !== 0`).

**Buffer size alignment** -- Correct. `Math.max(4, Math.ceil(rawSize / 4) * 4)` satisfies WebGPU's minimum buffer size of 4 bytes and 4-byte alignment requirement.

**Padding for non-aligned writes** -- Correct. The fast path avoids allocation when data is already aligned. The slow path copies into a zero-padded `Uint8Array` to prevent reading past the source `ArrayBuffer` boundary, which would cause a WebGPU validation error.

**dstOffset parameter** -- Correct. Passed directly to `GPUQueue.writeBuffer` as `bufferOffset`. Default `0` preserves backward compatibility. Type signature in `Gfx` interface matches implementation.

### API Consistency with Sokol

Matches `sg_update_buffer` semantics well. Sokol's C API also validates immutable buffer updates at runtime. The `dstOffset` parameter is an extension beyond Sokol's C API (`sg_update_buffer` always writes from offset 0), but it's a natural WebGPU affordance and the optional parameter with default 0 keeps the base API shape identical.

### Type Safety

- `updateBuffer` signature in `Gfx` interface correctly uses `dstOffset?: number`
- Implementation default `dstOffset = 0` matches the optional interface declaration
- Type-checks cleanly with `tsc --noEmit` (verified)

1. **No `dstOffset` alignment validation**: WebGPU requires `bufferOffset` to be a multiple of 4. The implementation relies on WebGPU's own validation for this. A proactive check with a clearer error message could be helpful, but the WebGPU error is descriptive enough.

2. **No bounds check for `dstOffset + writeBytes > buffer.size`**: Again, WebGPU validates this, but a proactive check could improve DX. Non-blocking.

3. **`node_modules` in the repo**: The repo tracks `node_modules` on main -- this is a pre-existing repo-wide issue, not introduced by this PR.

### Type-check result

```
npx tsc --noEmit  # passes cleanly, no errors
```

PR #28, opened 2026-04-19
Implement device lost handling and configurable GPU requirements

### LINKED ISSUE #1: Core GPU device/context initialization and swapchain management
## Summary
Initialize WebGPU device, configure the canvas context, and manage the swapchain for frame presentation.

## Current State
`app.ts:run()` already handles adapter/device request, context configuration, and canvas format detection. `gfx.ts:createGfx()` receives the initialized device, canvas, context, and format.

## Requirements
- [x] Request GPU adapter and device (`app.ts:15-18`)
- [x] Configure canvas context with preferred format (`app.ts:19-20`)
- [x] Swapchain texture acquisition in `beginPass` (`gfx.ts:268`)
- [ ] Device lost handling (`device.lost` promise)
- [ ] Feature/limit querying and validation
- [ ] Configurable device requirements (features, limits) via `AppDesc`

## References
- Sokol: `sg_setup()`, `sg_shutdown()`
- WebGPU: `navigator.gpu.requestAdapter()`, `GPUDevice`

rjwalters:
Closes #1

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -12,10 +12,23 @@ export async function run(desc: AppDesc): Promise<() => void> {
     throw new Error("WebGPU is not supported in this browser");
   }
 
-  const adapter = await navigator.gpu.requestAdapter();
+  const adapterOptions: GPURequestAdapterOptions = {};
+  if (desc.powerPreference !== undefined) {
+    adapterOptions.powerPreference = desc.powerPreference;
+  }
+  const adapter = await navigator.gpu.requestAdapter(adapterOptions);
   if (!adapter) throw new Error("Failed to get GPU adapter");
 
-  const device = await adapter.requestDevice();
+  const reqFeatures: GPUFeatureName[] = desc.requiredFeatures ?? [];
+  const unsupported = reqFeatures.filter(f => !adapter.features.has(f));
+  if (unsupported.length > 0) {
+    throw new Error(`GPU adapter does not support required features: ${unsupported.join(", ")}`);
+  }
+
+  const device = await adapter.requestDevice({
+    requiredFeatures: reqFeatures,
+    requiredLimits: desc.requiredLimits,
+  });
   const context = canvas.getContext("webgpu")!;
   const format = navigator.gpu.getPreferredCanvasFormat();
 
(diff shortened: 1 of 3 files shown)


rjwalters:
## Loom Judge Review — PR #28

**Verdict: APPROVE**

### Type Check
TypeScript (`tsc --noEmit`) passes cleanly on the PR-28 branch code. The one pre-existing error at `src/gfx.ts:130` (SharedArrayBuffer/ArrayBuffer mismatch in `writeTexture`) is unchanged by this PR and exists on `main` as well — not introduced here.

### Correctness

**`device.lost` handler (app.ts)**
- Handler is attached _before_ `await desc.init(gfx)`, which correctly closes the race where device loss could occur during async init. The comment explains this intent clearly.
- Setting `running = false` inside the handler stops the RAF loop, which is the right behavior.
- Calling `desc.deviceLost?.(info.reason, info.message)` gives callers an escape hatch without making it required. Correct use of optional chaining.
- The `running` flag is hoisted above the handler attachment, so the handler's closure captures it correctly.

**Adapter options (app.ts)**
- `powerPreference` is conditionally added only when defined — avoids sending `undefined` as the field value to the WebGPU API, which is correct.
- `requiredFeatures` pre-validation against `adapter.features` before calling `requestDevice` gives a cleaner error message than the browser's opaque rejection. Good defensive practice.
- `requiredLimits` is passed through directly as `Record<string, number>` — matches the WebGPU spec's `GPUSupportedLimits` shape for `requestDevice`.

**Uniform buffer limit check (gfx.ts)**
- Guard at `createGfx` entry: `UNIFORM_BUFFER_SIZE > device.limits.maxUniformBufferBindingSize` — throws early with a descriptive message instead of silently creating an oversized buffer. Correct placement (after device is obtained, before buffer allocation).

### Type Safety
- `GPURequestAdapterOptions`, `GPUFeatureName[]`, `GPUDeviceLostReason` — all proper WebGPU API types, no `any` casts introduced.
- `powerPreference?: "low-power" | "high-performance"` in `AppDesc` is a string literal union, not a raw `string`. This is slightly narrower than `GPUPowerPreference` (which is the same union), so it's fine but could reference the WebGPU type directly for future-proofing. Minor style note only, not a blocking issue.
- `requiredLimits?: Record<string, number>` is a reasonable approximation of the WebGPU limit object shape.

### WebGPU API Usage
All API calls are correct per spec:
- `navigator.gpu.requestAdapter(options)` — correct signature
- `adapter.features.has(f)` — correct way to query adapter features
- `adapter.requestDevice({ requiredFeatures, requiredLimits })` — correct
- `device.lost` is a `Promise<GPUDeviceLostInfo>`, and `.then()` is the correct way to attach a handler without blocking execution
- `device.limits.maxUniformBufferBindingSize` — correct limit property name

### Summary
The implementation is clean, correct, and follows WebGPU best practices. No issues blocking approval.

PR #29, opened 2026-04-19
Add pause/resume, error boundary, frame hooks and rate limiting

### LINKED ISSUE #12: App lifecycle (init/frame/cleanup) with async WebGPU setup
## Summary
Manage application lifecycle with async initialization (WebGPU setup), per-frame callbacks, and cleanup.

## Current State
`app.ts:run()` handles the full lifecycle: async GPU init, `init` callback, `requestAnimationFrame` loop calling `frame`, and returns a cleanup function.

## Requirements
- [x] Async WebGPU initialization (`app.ts:14-20`)
- [x] Init callback with Gfx context (`app.ts:39`)
- [x] Frame loop via `requestAnimationFrame` (`app.ts:85-90`)
- [x] Cleanup function returned (`app.ts:94-103`)
- [ ] Pause/resume support (e.g., when tab is hidden via `visibilitychange`)
- [ ] Error boundary — catch and report errors in frame callback
- [ ] Pre-frame and post-frame hooks
- [ ] Configurable target frame rate / frame limiting

## References
- Sokol: `sapp_run()`, `sapp_desc` callbacks
- WebGPU: `requestAnimationFrame`

rjwalters:
Closes #12

### `src/types.ts`
Added four optional fields to `AppDesc`:
- `onError?: (err: unknown) => void` — custom error handler for frame exceptions
- `preFrame?: (gfx: Gfx) => void` — called before `frame` each tick
- `postFrame?: (gfx: Gfx) => void` — called after `frame` each tick (even on error)
- `targetFps?: number` — cap the frame rate; undefined = uncapped (existing behaviour)

### `src/app.ts`
Rewrote the frame loop (was 7 lines, now ~35) to add:
- **Pause/resume**: `paused` boolean toggled by a `visibilitychange` listener registered via `eventMap` so it is automatically removed on cleanup. Re-arms `requestAnimationFrame` when the tab becomes visible again.
- **Error boundary**: `desc.frame(gfx)` is wrapped in try/catch. On error, `postFrame` fires, then `onError` is called if provided; otherwise the error is logged and the loop stops.
- **Pre/post hooks**: `desc.preFrame?.(gfx)` fires before the frame call; `desc.postFrame?.(gfx)` fires after (and also on error path before returning).
- **Frame limiting**: `targetInterval` computed from `targetFps`; each rAF callback compares the incoming `DOMHighResTimeStamp` against `lastFrameMs` and skips work (but reschedules) until enough time has elapsed.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -173,7 +173,11 @@ export async function run(desc: AppDesc): Promise<() => void> {
     } catch (err) {
       desc.postFrame?.(gfx);
       if (desc.onError) {
-        desc.onError(err);
+        if (desc.onError(err) === false) {
+          running = false;
+        } else {
+          requestAnimationFrame(frame);
+        }
       } else {
         console.error("[sokol-ts] frame error:", err);
         running = false;
--- a/src/app.ts
+++ b/src/app.ts
@@ -147,10 +147,40 @@ export async function run(desc: AppDesc): Promise<() => void> {
   resizeObserver.observe(canvas);
 
   // Frame loop
-  function frame() {
-    if (!running) return;
+  let paused = false;
+
+  const onVisibilityChange = () => {
+    paused = document.hidden;
+    if (!paused && running) requestAnimationFrame(frame);
+  };
+  document.addEventListener("visibilitychange", onVisibilityChange);
+  eventMap.push([document, "visibilitychange", onVisibilityChange]);
+
+  const targetInterval = desc.targetFps ? 1000 / desc.targetFps : 0;
+  let lastFrameMs = 0;
+
+  function frame(timestampMs: number) {
+    if (!running || paused) return;
+    if (targetInterval > 0 && timestampMs - lastFrameMs < targetInterval) {
+      requestAnimationFrame(frame);
+      return;
+    }
+    lastFrameMs = timestampMs;
     resize();
-    desc.frame(gfx);
+    desc.preFrame?.(gfx);
+    try {
+      desc.frame(gfx);
+    } catch (err) {
+      desc.postFrame?.(gfx);
+      if (desc.onError) {
+        desc.onError(err);
+      } else {
+        console.error("[sokol-ts] frame error:", err);
+        running = false;
+      }
+      return;
+    }
+    desc.postFrame?.(gfx);
     requestAnimationFrame(frame);
   }
   requestAnimationFrame(frame);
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review

**Type-check:** `tsc --noEmit` passes cleanly. No type errors.

### Summary

The implementation is well-structured and correctly integrates all four features into the existing frame loop. The visibility-change listener is correctly registered via `eventMap` so it is removed on cleanup (the direct `document.addEventListener` + manual `eventMap.push` pattern is necessary because the `listen()` helper's type constraint is `K extends keyof HTMLElementEventMap`, which excludes `Document`).

### Issue: `onError` stops the loop with no opt-out

When `onError` is provided and a frame error is caught, the code does:

```ts
desc.postFrame?.(gfx);
if (desc.onError) {
  desc.onError(err);
} else {
  console.error("[sokol-ts] frame error:", err);
  running = false;
}
return;  // ← loop never re-arms
```

In both branches the loop terminates — the `return` before `requestAnimationFrame(frame)` is unconditional. The default (no `onError`) case is intentional: log and stop. But a caller who supplies `onError` almost certainly wants to *handle* the error and continue the loop (e.g. skip one bad frame, log to a UI, recover state). There is no way to achieve this today.

**Suggested fix** — let `onError` return a boolean to signal whether to continue:

```ts
// types.ts
onError?: (err: unknown) => boolean | void;

// app.ts — in the catch block
const shouldContinue = desc.onError ? desc.onError(err) !== false : false;
if (!shouldContinue) {
  running = false;
  return;
}
// fall through to requestAnimationFrame(frame)
```

Or, at minimum, always re-arm after `onError` (opt-in to stopping via the cleanup function the caller already holds), and document that the default no-handler path stops the loop.

### Minor observation

`preFrame` is not called in the error branch (the throw happens inside the `try` after `preFrame` already fired). This is consistent with the PR description ("postFrame fires even on error") and seems intentional — just worth calling out in a doc comment so future readers understand the contract.

### Verdict

Changes requested on one actionable item: the `onError` loop-termination behavior needs either a continuation signal in the callback or at minimum explicit documentation of the stop-always semantics.

rjwalters:
## Loom Judge Review

**Type-check:** `tsc --noEmit` passes (the only error is a pre-existing `gfx.ts:125` `SharedArrayBuffer` issue on `main` — not introduced by this PR).

### Fix verification

The reported issue was that `onError` unconditionally terminated the frame loop. The fix is correct:

- `onError` now returns `boolean | void` (types.ts)
- `onError(err) === false` → stops the loop (`running = false`)
- Any other return (`true`, `undefined`/void) → re-arms `requestAnimationFrame` and the loop continues
- No `onError` provided → logs and stops (unchanged, correct default)

Both branches of the `if (desc.onError)` block handle their own continuation/termination, and the trailing `return` correctly skips the normal-path `postFrame` + `requestAnimationFrame`.

### No new issues found

The PR is clean. Approving.

PR #30, opened 2026-04-19
Add high-precision timing module (sokol_time equivalent)

### LINKED ISSUE #21: sokol_time equivalent — high-precision timing
## Summary
High-precision timing utilities for frame timing, elapsed time, and time-based animation.

## Current State
`gfx.ts:commit()` computes `dt` using `performance.now()` (`gfx.ts:346-348`). Frame count is tracked.

## Requirements
- [x] Delta time per frame (`gfx.ts:dt`)
- [x] Frame counter (`gfx.ts:frameCount`)
- [ ] Total elapsed time since init
- [ ] `stm_now()` equivalent — high-precision timestamp
- [ ] `stm_since()` / `stm_diff()` — duration helpers
- [ ] `stm_laptime()` — lap timer for profiling
- [ ] Unit conversion helpers (sec, ms, us, ns)
- [ ] Moving average frame time for smoothed FPS display

## References
- Sokol: `stm_setup()`, `stm_now()`, `stm_diff()`, `stm_sec()`
- Web: `performance.now()`, `performance.timeOrigin`

rjwalters:
Closes #21

## Summary
- New `src/stm.ts` module with `Stm` interface and `createStm(fpsWindowSize?)` factory, mirroring `sokol_time.h` design
- All timestamps are `DOMHighResTimeStamp` values (ms) from `performance.now()`; unit helpers (`sec`, `ms`, `us`, `ns`) convert outward on read
- `fps()` uses an O(1) circular-buffer moving average (default 60-sample window) with epsilon clamping to handle zero/negative dt gracefully
- `src/index.ts` gains one export line: `export { createStm } from "./stm.js"`
- `npx tsc --noEmit` passes cleanly; no changes to `gfx.ts`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
 export { run } from "./app.js";
 export { createGfx } from "./gfx.js";
+export { createStm } from "./stm.js";
 export * from "./types.js";
--- a/src/stm.ts
+++ b/src/stm.ts
@@ -1,0 +1,123 @@
+// stm.ts — High-precision timing utilities (sokol_time equivalent)
+//
+// All timestamps are DOMHighResTimeStamp values in milliseconds (sub-millisecond
+// precision via performance.now()). Unit conversion helpers convert outward to
+// the caller's preferred unit, mirroring how sokol_time.h stores ticks and
+// converts on read.
+//
+// Usage:
+//   const stm = createStm();
+//   const tick = stm.now();
+//   // ... do work ...
+//   console.log(stm.ms(stm.since(tick)));     // ms elapsed
+//   console.log(stm.sec(stm.elapsed()));      // total seconds since init
+//   console.log(stm.fps(gfx.dt));             // smoothed frames per second
+
+export interface Stm {
+  // Snapshot the current high-precision clock. Returns an opaque tick value (ms).
+  now(): number;
+
+  // Duration between two ticks (absolute value, always >= 0).
+  diff(a: number, b: number): number;
+
+  // Duration from a past tick to right now (in ms).
+  since(startTick: number): number;
+
+  // Lap timer: returns duration since the last call to laptime() (or since
+  // createStm() was called on the very first invocation). Updates the internal
+  // lap-start reference on each call. Useful for per-section profiling.
+  laptime(prevRef?: { t: number }): number;
+
+  // Total elapsed time (in ms) since createStm() was called.
+  elapsed(): number;
+
+  // Unit conversion helpers — input is a duration in milliseconds.
+  sec(d: number): number;  // d / 1000
+  ms(d: number): number;   // identity — ms is the base unit
+  us(d: number): number;   // d * 1000
+  ns(d: number): number;   // d * 1_000_000
+
+  // Smoothed FPS using an N-sample circular-buffer moving average.
+  // Call once per frame passing the frame delta in seconds (e.g. gfx.dt).
+  // Returns frames-per-second. Clamps dtSeconds to a small epsilon to guard
+  // against zero/negative values on the first frame or after tab backgrounding.
+  fps(dtSeconds: number): number;
+}
+
+// Minimum dt accepted by fps() — prevents division by zero and absorbs
+// first-frame spikes caused by the gfx.ts lastFrameTime=0 initialisation.
+const MIN_DT = 1e-6; // 1 microsecond expressed in seconds
+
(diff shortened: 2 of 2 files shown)


rjwalters:
## Loom Judge Review — PR #30: High-precision timing module (stm.ts)

**Type-check:** `npx tsc --noEmit` passes cleanly. No type errors.

---

### API Design

The `Stm` interface cleanly mirrors `sokol_time.h` conventions: opaque tick values, unit conversion helpers that convert outward on read, and a factory function (`createStm`) following the established `createGfx` pattern in this codebase. The module is self-contained with zero imports and no changes to `gfx.ts` — good separation of concerns.

**Strengths:**
- `diff(a, b)` returns `Math.abs(a - b)` — always non-negative, correct for a general-purpose duration utility.
- `since()` and `elapsed()` both clamp with `Math.max(0, ...)` — guards against monotonic clock weirdness on clock rollback (unlikely with `performance.now()` but correct to be defensive).
- `laptime()` dual-mode design (external `{ t: number }` ref object vs. internal `_lapStart`) is flexible without overcomplicating the interface. The `prevRef` pattern is idiomatic for multi-section profiling without requiring multiple `Stm` instances.
- The `Float64Array` circular buffer for FPS is an appropriate choice — avoids GC pressure in the hot frame path compared to a regular array of numbers.
- `_sum` accumulation with eviction (`_sum -= _window[_head]` before insertion) is the correct O(1) running-sum approach.

**One concern — `laptime()` API ambiguity:** The JSDoc says "returns duration since the last call to `laptime()` ... on the very first invocation" when no `prevRef` is supplied. However, callers using the internal state share a single `_lapStart` — if two different call sites both call `laptime()` without a `prevRef`, they silently interfere with each other. The recommended usage pattern (passing `prevRef`) avoids this, but the default no-arg form is a hidden footgun when code grows. A minor doc note or a usage caution would preempt confusion.

---

### Precision

- `DOMHighResTimeStamp` from `performance.now()` provides sub-millisecond resolution (typically microsecond on modern browsers). Storing in `Float64Array` preserves this — `Float64` has 53 bits of mantissa, sufficient for timestamps up to ~104 days without precision loss.
- Unit conversion helpers (`sec`, `ms`, `us`, `ns`) are pure arithmetic with no rounding — callers who need integer nanoseconds must round themselves, which is appropriate for a library.

---

### Edge Cases

- **FPS cold start:** `_count` starts at 0, `_sum` at 0. First call: `_sum -= _window[0]` (evicts 0), then `_window[0] = dt`, `_sum = dt`, `_count = 1`. Returns `1 / dt` — correct single-sample FPS. Good.
- **Tab backgrounding / large dt spike:** `MIN_DT = 1e-6` clamps extreme spikes upward but does **not** clamp large dt values *downward*. A single backgrounded-tab frame (dt = 5s) will pollute the rolling average for `fpsWindowSize` frames. This is a known trade-off (sokol_time.h has the same behavior), but worth acknowledging. If this is used in a game loop that calls `requestAnimationFrame`, the browser already throttles to ~1fps when backgrounded, so the spike magnitude is bounded. Acceptable for this scope.
- **`fpsWindowSize = 0`:** Would cause `NaN` (`_count / _sum` where `_count` stays 0 and `_sum` stays 0 — actually `0/0`). The parameter has a sensible default (60) and no guard for `<= 0`. A `fpsWindowSize < 1` guard or assertion would be good hygiene, but this is a minor issue.
- **`Float64Array` precision for `_sum`:** Accumulated addition/subtraction of many small floats can accumulate error. At 60 samples of ~16ms each, `_sum ≈ 1.0` — well within Float64 precision. Not a real issue in practice.

---

### Integration with `gfx.ts`

`gfx.dt` is computed in `commit()` as `(now - lastFrameTime) / 1000`, producing seconds. `stm.fps(gfx.dt)` will work correctly as documented. The first-frame issue (`lastFrameTime = 0`) means `gfx.dt` on frame 0 will be enormous (~seconds since page load), which `MIN_DT` does not mitigate — it clamps the *minimum* not the maximum. The FPS window will absorb this spike over 60 frames. The PR description correctly identifies this and documents it.

---

### Summary

This is a clean, well-reasoned implementation. The code is correct, the design is consistent with the codebase, the type-check is clean, and the edge cases are either handled or documented. The two minor items (no-arg `laptime()` footgun potential, no guard on `fpsWindowSize <= 0`) do not warrant blocking the PR — they are quality-of-life concerns, not correctness bugs.

**Verdict: Approve.**

PR #32, opened 2026-04-19
Add shader compilation error handling and combined WGSL support

### LINKED ISSUE #5: Shader module creation from WGSL source
## Summary
Create shader modules from WGSL source code, supporting separate vertex and fragment stages.

## Current State
`gfx.ts:makeShader` creates separate `GPUShaderModule` for vertex and fragment from WGSL source strings. Stores source for potential hot-reload.

## Requirements
- [x] Create vertex and fragment shader modules (`gfx.ts:149-155`)
- [x] Store source strings for recompilation (`gfx.ts:153`)
- [ ] Shader compilation error handling and reporting (`compilationInfo()`)
- [ ] Combined vertex+fragment in single WGSL module (auto-detect entry points)
- [ ] Shader reflection for automatic bind group layout generation
- [ ] Compute shader support

## References
- Sokol: `sg_make_shader()`
- WebGPU: `GPUDevice.createShaderModule()`

rjwalters:
Closes #5

## Changes

- **`makeShader` is now async** — awaits `getCompilationInfo()` on both shader modules before storing them. Throws a descriptive `Error` (with stage name, line, and column) on any `"error"` message; calls `console.warn` for `"warning"` messages.
- **Combined WGSL source** — `ShaderDesc` gains an optional `source` field. When provided, a single `GPUShaderModule` is created and reused for both vertex and fragment stages (compiled once, checked once).
- **Custom entry points** — `ShaderDesc` gains `vertexEntry` (default `"vs_main"`) and `fragmentEntry` (default `"fs_main"`). These are stored in `ShaderSlot` and threaded through `makePipeline` instead of the previous hardcoded strings.
- **`Gfx` interface updated** — `makeShader` signature is now `Promise<SgShader>`.

## Files changed

- `src/types.ts` — `ShaderDesc` fields updated; `Gfx.makeShader` return type changed
- `src/gfx.ts` — `ShaderSlot` fields updated; `makeShader` rewritten as async; `makePipeline` uses stored entry points

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -29,8 +29,8 @@ interface SamplerSlot {
 interface ShaderSlot {
   vertexModule: GPUShaderModule;
   fragmentModule: GPUShaderModule;
-  vertexSource: string;
-  fragmentSource: string;
+  vertexEntry: string;
+  fragmentEntry: string;
 }
 
 interface PipelineSlot {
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review — PR #32

**Verdict: APPROVED**

### Type Check
`npx tsc --noEmit` — **zero errors, zero warnings**.

---

### Evaluation

#### 1. `makeShader` is now async — correct
`async makeShader(desc: ShaderDesc): Promise<SgShader>` is properly declared both in the implementation (`gfx.ts`) and the `Gfx` interface (`types.ts`). The `await` on `Promise.all([...])` is correct and the handle `h` is only stored in the map after compilation succeeds, so a failed shader never ends up in the slot map.

#### 2. Compilation error handling — correct
`checkCompilation` is a well-scoped inner async function. It iterates `GPUCompilationInfo.messages`, throws a descriptive `Error` (stage, lineNum, linePos, message) for `"error"` type, and `console.warn`s for `"warning"` type. The `"info"` type is silently ignored — acceptable for a library.

#### 3. Combined WGSL (`source` field) — correct
When `desc.source` is provided, a single `GPUShaderModule` is created and reused as both `vertexModule` and `fragmentModule`. The identity check `fragmentModule !== vertexModule` correctly skips the redundant second `checkCompilation` call. When `source` is absent, `desc.vertexSource!` and `desc.fragmentSource!` are used with non-null assertions — safe given the documented contract. `ShaderDesc` fields are all optional at the type level; the runtime behavior relies on callers following the documented exclusive-or pattern (`source` xor `vertexSource`+`fragmentSource`), which is a reasonable API for this codebase.

#### 4. Custom entry points — correct
`vertexEntry` and `fragmentEntry` default to `"vs_main"` / `"fs_main"` respectively, preserving full backward compatibility. They are stored in `ShaderSlot` and threaded through `makePipeline` — both the vertex and fragment `entryPoint` fields now use `shd.vertexEntry` / `shd.fragmentEntry` instead of the previous hardcoded strings.

#### 5. `ShaderSlot` cleanup — correct
The old `vertexSource`/`fragmentSource` string fields are removed from `ShaderSlot`; they served no purpose post-construction since the compiled `GPUShaderModule` objects are what matter.

---

- **No validation guard for the `source`-xor-`vertexSource`+`fragmentSource` contract.** If a caller passes neither `source` nor `vertexSource`, the `!` assertion silently yields `undefined` as the WGSL code. A small runtime guard (`if (!combinedSource && !desc.vertexSource) throw new Error(...)`) would make misuse fail loudly. This is a hardening suggestion, not a correctness bug.
- **Label for combined module**: when `source` is used, both `_vs`-suffixed label and no `_fs` label are created. Acceptable — the single module does double duty.

These are style/robustness notes only and do not block approval.

---

**loom:pr** — ready to merge.

PR #33, opened 2026-04-19
Support multiple app instances with per-instance state

### LINKED ISSUE #15: Multiple simultaneous app instances on separate canvases
## Summary
Support running multiple independent sokol-ts app instances, each with its own canvas, device, and state.

## Current State
`run()` is a standalone function that creates a fresh Gfx context per call. The global `nextId` counter in `gfx.ts` is shared, but handle IDs don't collide across instances since each instance has its own resource maps.

## Requirements
- [x] Independent per-instance state (each `createGfx` call is self-contained)
- [ ] Per-instance ID counter (currently global `nextId` at `gfx.ts:9`)
- [ ] Shared device option (multiple canvases on same GPU device for efficiency)
- [ ] Independent cleanup without affecting other instances
- [ ] Cross-instance resource sharing (textures, buffers)
- [ ] Test with 2+ simultaneous instances

## References
- Sokol: single-instance design — this extends beyond Sokol's model
- WebGPU: multiple `GPUCanvasContext` on same `GPUDevice`

rjwalters:
Closes #15

## Changes

- **`src/gfx.ts`**: Moved `nextId` counter and `handle()` helper inside `createGfx()` closure. Each instance now counts independently from 1, eliminating the shared mutable module-level state.

- **`src/types.ts`**: Added `device?: GPUDevice` field to `AppDesc`. Callers can supply a pre-created device to share it across multiple `run()` instances.

- **`src/app.ts`**: When `desc.device` is provided, skip adapter/device creation. When cleaning up, only call `device.destroy()` if `run()` created the device — caller-supplied devices are not destroyed (caller owns lifetime).

## Notes on event isolation

`keydown`/`keyup` listeners are bound to `window` but are tracked in a per-closure `eventMap` array. Cleanup correctly removes only that instance's handlers, leaving other instances unaffected. No code change required.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -12,23 +12,26 @@ export async function run(desc: AppDesc): Promise<() => void> {
     throw new Error("WebGPU is not supported in this browser");
   }
 
-  const adapterOptions: GPURequestAdapterOptions = {};
-  if (desc.powerPreference !== undefined) {
-    adapterOptions.powerPreference = desc.powerPreference;
-  }
-  const adapter = await navigator.gpu.requestAdapter(adapterOptions);
-  if (!adapter) throw new Error("Failed to get GPU adapter");
+  const callerDevice = desc.device;
+  const device = callerDevice ?? await (async () => {
+    const adapterOptions: GPURequestAdapterOptions = {};
+    if (desc.powerPreference !== undefined) {
+      adapterOptions.powerPreference = desc.powerPreference;
+    }
+    const adapter = await navigator.gpu.requestAdapter(adapterOptions);
+    if (!adapter) throw new Error("Failed to get GPU adapter");
 
-  const reqFeatures: GPUFeatureName[] = desc.requiredFeatures ?? [];
-  const unsupported = reqFeatures.filter(f => !adapter.features.has(f));
-  if (unsupported.length > 0) {
-    throw new Error(`GPU adapter does not support required features: ${unsupported.join(", ")}`);
-  }
+    const reqFeatures: GPUFeatureName[] = desc.requiredFeatures ?? [];
+    const unsupported = reqFeatures.filter(f => !adapter.features.has(f));
+    if (unsupported.length > 0) {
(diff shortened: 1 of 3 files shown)


rjwalters:
## Loom Judge Review

**Verdict: APPROVE**

### Type-check
`npx tsc --noEmit` — clean, zero errors.

### Multi-instance state isolation

**`src/gfx.ts` — `nextId` counter**
Moving `nextId` and `handle()` inside `createGfx()` is correct. Each instance now has its own counter starting at 1. Handle IDs are scoped to the owning `Gfx` instance's Maps (`buffers`, `images`, `samplers`, `shaders`, `pipelines`), so cross-instance ID collisions cannot cause incorrect lookups — each Map is private to its closure. No issue here.

**`src/app.ts` — device sharing**
The `callerDevice` capture is clean: `const callerDevice = desc.device` at the top of `run()` freezes the optional value into a local before any async work. The teardown guard `if (!callerDevice) device.destroy()` correctly gates destruction on ownership. A caller-supplied device is never destroyed by `run()`, so it remains valid for sibling instances. This is the correct ownership model.

**`src/types.ts` — `AppDesc.device?`**
Optional field placement after `canvas` and before `init` is fine. No breaking change to existing callers (all existing `AppDesc` literals simply omit the field).

### Event isolation
`keydown`/`keyup` listeners are bound to `window` but captured in a per-closure `eventMap`. Cleanup iterates only that closure's `eventMap` and removes exactly those handlers. Other instances are unaffected. No code change was needed and none was made — correct.

### Cleanup safety
- If the caller supplies a device: `run()` never calls `device.destroy()`, so the caller retains full control of device lifetime.
- If `run()` creates the device: it is destroyed on cleanup as before.
- `resizeObserver.disconnect()` and all `eventMap` listeners are always cleaned up regardless of device ownership.
- The frame loop halts via `running = false` before any async teardown.

### No concerns
The change is minimal, well-scoped, and does not introduce any shared mutable state or race conditions. The IIFE for device creation (`await (async () => { ... })()`) is idiomatic and avoids any variable shadowing issues.

PR #34, opened 2026-04-19
Add generation-counted resource pools with handle validation

### LINKED ISSUE #10: Resource pool and handle validation
## Summary
Implement a proper resource pool with generation-counted handles for safe resource management and dangling handle detection.

## Current State
Resources are stored in `Map<number, Slot>` collections with a simple incrementing ID counter (`gfx.ts:9-12`). No generation counting or handle validation.

## Requirements
- [x] Basic resource storage via Maps (`gfx.ts:48-52`)
- [x] Unique handle generation (`gfx.ts:9-12`)
- [ ] Generation-counted handles (detect use-after-destroy)
- [ ] Fixed-size pool with slot reuse (matching Sokol's pool design)
- [ ] Handle validation on every API call
- [ ] Resource leak detection (warn on shutdown if resources not destroyed)
- [ ] `isValid(handle)` query function
- [ ] Pool exhaustion handling
- [ ] Debug labels for all resources (queryable)

## References
- Sokol: `_sg_pool_t`, `sg_resource_state`, handle generation bits
- WebGPU: N/A (this is a library-level concern)

rjwalters:
Closes #10

## Summary

- Replaces five `Map<number, Slot>` collections in `src/gfx.ts` with a generic `Pool<T>` class that packs a 16-bit slot index and 16-bit generation counter into each handle `id`, matching Sokol's `_sg_slot_t` strategy
- Slot 0 is reserved as a null sentinel; `Pool.get()` validates the generation on every lookup, catching use-after-destroy and stale-handle bugs that previously silently did nothing
- Adds `isValid(handle: Handle): boolean` to the `Gfx` interface and `shutdown(): void` for leak detection — `shutdown()` destroys any remaining GPU resources and emits a `console.warn` listing counts and debug labels of leaked resources
- `src/app.ts` now calls `gfx.shutdown()` between `desc.cleanup?.(gfx)` and `device.destroy()`

## Test plan

- [ ] Basic round-trip — create a buffer, `isValid()` → `true`; destroy it, `isValid()` → `false`
- [ ] Stale handle detection — destroy a buffer, re-create another (reuses slot index), use old handle in `updateBuffer()` → must throw
- [ ] Pool exhaustion — allocate `capacity + 1` buffers → throws descriptive error
- [ ] Slot reuse — destroy and re-create; new handle shares slot index but has incremented generation
- [ ] Leak detection — create resources, call `shutdown()` without destroying → `console.warn` fires with non-zero count
- [ ] Clean shutdown — destroy all before `shutdown()` → no warning
- [ ] `applyPipeline` with stale handle → throws "Invalid or stale pipeline handle"
- [ ] `applyBindings` with stale buffer → throws
- [ ] Debug label in leak warning — labeled buffer appears in warning message

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -334,6 +334,7 @@ export async function run(desc: AppDesc): Promise<() => void> {
     }
     desc.cleanup?.(gfx);
     context.unconfigure();
+    gfx.shutdown();
     if (!callerDevice) device.destroy();
   };
 }
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -1,10 +1,101 @@
 import {
   type SgBuffer, type SgImage, type SgSampler, type SgShader, type SgPipeline,
   type BufferDesc, type ImageDesc, type SamplerDesc, type ShaderDesc, type PipelineDesc,
-  type Bindings, type PassDesc, type Gfx, type DrawStats, type ShaderRecompileResult,
+  type Bindings, type PassDesc, type Gfx, type Handle, type DrawStats, type ShaderRecompileResult,
   BufferUsage, IndexType, LoadAction, StoreAction, PixelFormat, PrimitiveType, CullMode, CompareFunc,
   FilterMode, WrapMode,
 } from "./types.js";
+
+// ---------------------------------------------------------------------------
+// Generation-counted handle encoding (matches Sokol's _sg_slot_t strategy)
+// ---------------------------------------------------------------------------
+
+const POOL_BITS = 16;
+const GEN_BITS  = 16;
+const POOL_MASK = (1 << POOL_BITS) - 1;   // 0x0000_FFFF
+const GEN_MASK  = (1 << GEN_BITS)  - 1;   // 0x0000_FFFF
+const GEN_SHIFT = POOL_BITS;
+
+function encodeHandle(index: number, gen: number): number {
+  return ((gen & GEN_MASK) << GEN_SHIFT) | (index & POOL_MASK);
+}
+function slotIndex(id: number): number  { return id & POOL_MASK; }
+function generation(id: number): number { return (id >>> GEN_SHIFT) & GEN_MASK; }
+
+// ---------------------------------------------------------------------------
+// Pool implementation
+// ---------------------------------------------------------------------------
+
+const enum SlotState { INVALID = 0, ALLOC = 1, VALID = 2 }
+
+interface PoolSlot<T> {
+  gen:   number;
+  state: SlotState;
+  res:   T | undefined;
+}
+
+class Pool<T> {
+  private slots: PoolSlot<T>[];
+  private freeList: number[];
+
+  constructor(capacity: number) {
+    // slot 0 is reserved as the "null" sentinel
+    this.slots = Array.from({ length: capacity + 1 }, () =>
+      ({ gen: 0, state: SlotState.INVALID, res: undefined }));
+    this.freeList = Array.from({ length: capacity }, (_, i) => i + 1);
+  }
+
+  alloc(): number {
+    const index = this.freeList.pop();
+    if (index === undefined) throw new Error("Pool exhausted");
+    const slot = this.slots[index];
+    slot.gen = (slot.gen + 1) & GEN_MASK;   // increment; wraps within 16-bit generation range
+    slot.state = SlotState.ALLOC;
+    return encodeHandle(index, slot.gen);
+  }
+
+  set(id: number, res: T): void {
+    const s = this.lookup(id);
+    if (!s) throw new Error("Pool.set on invalid handle");
+    s.res = res;
+    s.state = SlotState.VALID;
+  }
+
+  get(id: number): T | undefined {
+    return this.lookup(id)?.res;
+  }
+
+  free(id: number): T | undefined {
+    const index = slotIndex(id);
+    const slot = this.slots[index];
+    if (slot.state === SlotState.INVALID || slot.gen !== generation(id)) return undefined;
+    const res = slot.res;
+    slot.res = undefined;
+    slot.state = SlotState.INVALID;
+    this.freeList.push(index);
+    return res;
+  }
+
+  private lookup(id: number): PoolSlot<T> | undefined {
+    const index = slotIndex(id);
+    if (index === 0) return undefined;
+    const slot = this.slots[index];
+    if (slot.state === SlotState.INVALID) return undefined;
+    if (slot.gen !== generation(id)) return undefined;  // stale generation
+    return slot;
+  }
+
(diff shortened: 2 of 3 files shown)


rjwalters:
## Loom Judge Review

**Type-check:** `npx tsc --noEmit` — clean, zero errors.

**Overall:** The architecture is solid — generation-counted bit-packing, slot-0 null sentinel, ALLOC→VALID two-phase construction, free-list, `isValid()`, and `shutdown()` leak detection all land correctly. Two must-fix bugs and three minor issues block approval.

---

### BUG 1 — Generation counter overflows and permanently breaks slots after 65 535 recycles

`Pool.alloc()` line 52:
```ts
slot.gen = (slot.gen + 1) | 0;   // increment; wraps safely with |0
```
`|0` converts to a signed 32-bit integer, so `slot.gen` grows past `GEN_MASK` (0xFFFF). After the 65 536th recycle of a slot, `slot.gen = 65536`, but `encodeHandle` encodes only `gen & GEN_MASK = 0`. The `lookup()` check `slot.gen !== generation(id)` then evaluates `65536 !== 0` → stale, and every subsequent `set()` throws `"Pool.set on invalid handle"`. The slot is permanently unusable without being removed from the free list.

**Fix:** mask the counter to stay within 16 bits:
```ts
slot.gen = (slot.gen + 1) & GEN_MASK;
```
(The comment "wraps safely with |0" is incorrect for a 16-bit field.)

---

### BUG 2 — `makePipeline` leaks a pool slot when the shader handle is invalid

```ts
makePipeline(desc: PipelineDesc): SgPipeline {
  const id = pipelinePool.alloc();          // slot consumed, state = ALLOC
  const shd = shaderPool.get(desc.shader.id);
  if (!shd) throw new Error("...");         // error thrown — slot never freed
```

`pipelinePool.alloc()` increments the generation, marks the slot `ALLOC`, and removes it from the free list. If `shaderPool.get()` returns `undefined`, the error is thrown and the ALLOC slot is stranded — it is never transitioned back to `INVALID` and never returned to the free list. Each call with a bad shader handle silently shrinks the pipeline pool by one.

The same latent pattern exists in all `make*()` functions if `device.createBuffer()` / `device.createTexture()` etc. throw (e.g. OOM), but the shader-validation case is the only one reachable via normal API misuse.

**Fix:** validate before allocating:
```ts
makePipeline(desc: PipelineDesc): SgPipeline {
  const shd = shaderPool.get(desc.shader.id);
  if (!shd) throw new Error("Invalid or stale shader handle");
  const id = pipelinePool.alloc();
  // ... rest unchanged
```

---

### ISSUE 3 — `applyPipeline` conflates two distinct error conditions

```ts
if (!slot || !passEncoder) throw new Error("Invalid or stale pipeline handle");
```

When `passEncoder` is null (no active pass), the message says "Invalid or stale pipeline handle", which is misleading and will make debugging painful. Split into two checks:
```ts
if (!slot) throw new Error("Invalid or stale pipeline handle");
if (!passEncoder) throw new Error("applyPipeline called outside of a pass");
```

---

### ISSUE 4 — `shutdown()` omits sampler labels from the leak report

This PR adds `desc: SamplerDesc` to `SamplerSlot` specifically to support label reporting, but the label-collection loop in `shutdown()` never reads it:
```ts
for (const s of leakedBuffers)   { if (s.desc.label) labels.push(s.desc.label); }
for (const s of leakedImages)    { if (s.desc.label) labels.push(s.desc.label); }
for (const s of leakedPipelines) { if (s.desc.label) labels.push(s.desc.label); }
// leakedSamplers is collected but its labels are never pushed
```
Add:
```ts
for (const s of leakedSamplers)  { if (s.desc.label) labels.push(s.desc.label); }
```

---

### ISSUE 5 — `ShaderSlot` stores no label, leaked shaders are anonymous in the report

`ShaderDesc` has a `label?: string` field, but `ShaderSlot` does not store it. Leaked shaders appear in the count but contribute no name to the label list. Storing `desc: ShaderDesc` (or just `label?: string`) in `ShaderSlot` and including it in the loop would make the warning complete.

---

**Verdict:** requesting changes on BUG 1 and BUG 2. Issues 3–5 are non-blocking but worth fixing in the same pass since the surface area is small.

rjwalters:
## Review: Changes Not Found

The PR branch (`worktree-agent-a230b996`) contains only the original commit (`307e0ac`). **Neither bug fix appears to have been pushed.** Both previously reported issues remain in the code:

### Bug 1: Generation counter still uses `| 0` (32-bit wrap)

In `Pool.alloc()`:
```ts
slot.gen = (slot.gen + 1) | 0;   // increment; wraps safely with |0
```

This wraps at 32 bits, but `encodeHandle` masks the generation to 16 bits (`gen & GEN_MASK`). After 65,536 recycles of a single slot:

- `slot.gen` becomes `65536` (0x10000)
- `encodeHandle` encodes generation as `65536 & 0xFFFF = 0`
- `generation(id)` extracts `0` from the handle
- `slot.gen !== generation(id)` evaluates to `65536 !== 0` → **lookup fails, handle appears stale**

**Fix:** `slot.gen = (slot.gen + 1) & GEN_MASK;`

### Bug 2: `pipelinePool.alloc()` still called before shader validation

In `makePipeline()`:
```ts
const id = pipelinePool.alloc();          // slot allocated
const shd = shaderPool.get(desc.shader.id);
if (!shd) throw new Error("...");         // throws without freeing `id`
```

If the shader handle is invalid, the allocated pipeline slot is leaked (never returned to the free list).

**Fix:** Either validate the shader *before* calling `alloc()`, or add a try/catch that calls `pipelinePool.free(id)` on error.

### Minor: `shutdown()` double-counts live resources

`shutdown()` calls `bufferPool.liveResources()` to destroy GPU objects, then calls it again to count leaks. Since the pool slots are never freed (only the GPU handles are destroyed), the second call returns the same list, so the leak warning always fires even though the GPU resources were just cleaned up. Consider freeing pool slots after destroying GPU resources, or capturing the list once and reusing it.

---

Please push the fixes and request re-review.

rjwalters:
**Loom Judge: Approved**

Verified both fixes in commit 822b5a0:

1. **Generation counter overflow** (`& GEN_MASK` instead of `| 0`): Confirmed. The `| 0` coercion to signed 32-bit could produce negative generation values that would never match the 16-bit mask applied in `generation()`. Using `& GEN_MASK` correctly wraps the counter within the 16-bit range.

2. **Pipeline slot leak** (shader validation before `pipelinePool.alloc()`): Confirmed. The shader lookup and validation now occur before `pipelinePool.alloc()`, so if the shader handle is invalid, no pool slot is allocated and subsequently leaked.

Type-check (`npx tsc --noEmit`) passes cleanly with no errors.

No other issues found.

PR #35, opened 2026-04-19
Fix DPI handling: change detection, touch coords, CSS size

### LINKED ISSUE #14: High-DPI / devicePixelRatio handling
## Summary
Handle high-DPI displays by scaling the canvas backing store to match the device pixel ratio.

## Current State
`app.ts` reads `devicePixelRatio` (configurable via `AppDesc.pixelRatio`) and scales the canvas width/height accordingly in the `resize()` function.

## Requirements
- [x] Apply devicePixelRatio to canvas dimensions (`app.ts:24-33`)
- [x] Configurable pixel ratio override (`app.ts:24`)
- [x] Scale mouse coordinates by pixel ratio (`app.ts:58-60`)
- [ ] Respond to DPI changes (e.g., moving window between monitors via `matchMedia`)
- [ ] Expose CSS vs physical size distinction in Gfx interface
- [ ] Touch coordinate scaling (currently uses `clientX/Y` which may not account for canvas offset)
- [ ] Optional DPI-independent coordinate mode

## References
- Sokol: `sapp_dpi_scale()`, `sapp_high_dpi()`
- Web: `window.devicePixelRatio`, `matchMedia('resolution')`

rjwalters:
Closes #14

## Changes

- **DPI change detection** (`src/app.ts`): replaced static `pixelRatio` capture with a `matchMedia('(resolution: Ndppx)')` listener using `{ once: true }` re-registration so the canvas rescales on every monitor DPI transition. Teardown removes the listener.

- **CSS size getters** (`src/types.ts`, `src/gfx.ts`): added `cssWidth`, `cssHeight`, and `dpiScale` to the `Gfx` interface and implemented them as getters on the `gfx` object (`canvas.clientWidth`, `canvas.clientHeight`, and derived ratio).

- **Touch coordinate fix** (`src/app.ts`): replaced raw `t.clientX * pixelRatio` with a `toCanvasTouch()` helper that subtracts `getBoundingClientRect()` to produce canvas-relative coordinates, matching mouse `offsetX/Y` behaviour.

- **`dpiIndependentCoords`** (`src/types.ts`, `src/app.ts`): added optional `AppDesc.dpiIndependentCoords` flag. When `true`, a `coordScale()` helper returns `1` instead of `pixelRatio`, delivering CSS-pixel coordinates for all mouse/touch events and `RESIZE` width/height.


rjwalters:
## Loom Judge Review — PR #35: Fix DPI handling

**TypeScript:** `npx tsc --noEmit` — clean, zero errors.

---

### Evaluation by area

#### 1. matchMedia DPI listener

The pattern is correct and idiomatic. Using `{ once: true }` and re-registering after each change handles cascading DPI transitions (e.g. moving across two external monitors with different pixel densities). The `desc.pixelRatio` guard in `onDpiChange` correctly short-circuits when the caller has overridden the ratio. The teardown in the returned cleanup function removes the listener and nulls the reference — no leak.

One nuance worth noting: `onDpiChange` references `dispatch` (defined later in the function body), but because `onDpiChange` is only ever invoked asynchronously from a DOM event — never synchronously during setup — hoisting is not a hazard here. Safe.

#### 2. Touch coordinate fix (`toCanvasTouch`)

Replacing `t.clientX * pixelRatio` with `(t.clientX - rect.left) * coordScale()` is the correct fix. `clientX/Y` are viewport-relative; subtracting `getBoundingClientRect().left/top` makes them canvas-relative, matching what `offsetX/Y` gives mouse events. The `getBoundingClientRect()` call-per-touch-event has negligible cost (it's a layout-cached value on a static canvas). No issues.

#### 3. CSS size getters (`cssWidth`, `cssHeight`, `dpiScale`)

- `canvas.clientWidth / clientHeight` — correct CSS-pixel values; consistent with what `resize()` already reads.
- `dpiScale` computed as `canvas.width / (canvas.clientWidth || 1)` — the `|| 1` guard prevents a divide-by-zero when the canvas is temporarily zero-sized (display:none, zero-area layout). This is derived from canvas state rather than the local `pixelRatio` variable, which is the right choice: it reflects the *actual* current ratio on the GPU canvas, not the stale `pixelRatio` from when the last `onDpiChange` fired.

#### 4. `dpiIndependentCoords` flag

Clean opt-in design. `coordScale()` is a minimal closure returning `1 | pixelRatio`; called at event-dispatch time so it always reads the live `pixelRatio` after a DPI change. RESIZE events from both `ResizeObserver` and `onDpiChange` consistently check the flag and dispatch `clientWidth/clientHeight` vs `canvas.width/canvas.height`. The two code paths are symmetric — no drift risk.

---

- `coordScale` is defined as `const coordScale = () => ...` (arrow function) rather than a `function` declaration, which is fine but slightly inconsistent with the rest of the module's style. Not a blocker.
- Mouse `deltaX/deltaY` on `mousemove` is not scaled by `coordScale()`, which is correct — movement deltas are already in CSS pixels from the browser and scaling them would be wrong. Good.

---

**Verdict: APPROVE.** All four targeted areas are implemented correctly. TypeScript is clean. No regressions spotted.

PR #36, opened 2026-04-19
Add indirect draw, validation, auto element count, and draw stats

### LINKED ISSUE #9: Draw call dispatch (indexed + non-indexed + instanced)
## Summary
Issue draw calls supporting indexed, non-indexed, and instanced rendering.

## Current State
`gfx.ts:draw` dispatches either `drawIndexed` or `draw` based on the current pipeline's index type. Supports instance count.

## Requirements
- [x] Non-indexed draw (`gfx.ts:328`)
- [x] Indexed draw (`gfx.ts:326`)
- [x] Instanced rendering (`numInstances` parameter, `gfx.ts:322`)
- [ ] Indirect draw calls (`drawIndirect`, `drawIndexedIndirect`)
- [ ] Draw call validation (element count vs buffer size)
- [ ] Automatic vertex/index count from buffer size
- [ ] Draw call statistics/profiling

## References
- Sokol: `sg_draw()`
- WebGPU: `GPURenderPassEncoder.draw()`, `.drawIndexed()`

rjwalters:
Closes #9

## Changes

- **`BufferUsage.INDIRECT`** added to enum; `gpuBufferUsage()` maps it to `GPUBufferUsage.INDIRECT`
- **`drawIndirect(indirectBuffer, indirectOffset?)`** dispatches `drawIndexedIndirect` or `drawIndirect` based on the active pipeline's `indexType`
- **Bound buffer caching** in `applyBindings()` stores `BufferSlot` refs for vertex and index buffers into `boundVertexBuffers` / `boundIndexBuffer`
- **Draw validation** in `draw()` checks that `baseElement + numElements` does not exceed the buffer's element capacity (throws a descriptive `Error` before recording the GPU command)
- **Auto element count** — `numElements` is now optional; when omitted or `0`, count is derived from `bufferSize / bytesPerIndex` (indexed) or `bufferSize / stride` (non-indexed) minus `baseElement`
- **`DrawStats` interface** with `drawCalls`, `totalElements`, `indirectDrawCalls`; exposed as `gfx.frameStats` (returns a snapshot copy); accumulated in `draw`/`drawIndirect`, reset on `beginPass`

## Files changed

| File | Change |
|---|---|
| `src/types.ts` | `BufferUsage.INDIRECT`, `DrawStats` interface, `drawIndirect()` + `frameStats` on `Gfx`, `numElements` made optional |
| `src/gfx.ts` | All implementation wiring described above |

`npx tsc --noEmit` passes cleanly.

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -1,7 +1,7 @@
 import {
   type SgBuffer, type SgImage, type SgSampler, type SgShader, type SgPipeline,
   type BufferDesc, type ImageDesc, type SamplerDesc, type ShaderDesc, type PipelineDesc,
-  type Bindings, type PassDesc, type Gfx,
+  type Bindings, type PassDesc, type Gfx, type DrawStats,
   BufferUsage, IndexType, LoadAction, PixelFormat, PrimitiveType, CullMode, CompareFunc,
   FilterMode, WrapMode,
 } from "./types.js";
@@ -61,6 +61,9 @@ export function createGfx(
   let uniformOffset = 0;
   let uniformBuffer: GPUBuffer | null = null;
   let uniformBindGroup: GPUBindGroup | null = null;
+  let boundVertexBuffers: BufferSlot[] = [];
+  let boundIndexBuffer: BufferSlot | null = null;
+  let _frameStats: DrawStats = { drawCalls: 0, totalElements: 0, indirectDrawCalls: 0 };
 
   const UNIFORM_BUFFER_SIZE = 65536; // 64KB uniform staging
 
@@ -77,7 +80,10 @@ export function createGfx(
   function gpuBufferUsage(usage: BufferUsage | undefined): number {
     const base = GPUBufferUsage.COPY_DST;
     switch (usage) {
-      default: return base | GPUBufferUsage.VERTEX | GPUBufferUsage.INDEX;
+      case BufferUsage.INDIRECT:
+        return base | GPUBufferUsage.INDIRECT;
+      default:
+        return base | GPUBufferUsage.VERTEX | GPUBufferUsage.INDEX;
     }
   }
 
@@ -88,6 +94,7 @@ export function createGfx(
     get height() { return canvas.height; },
     get dt() { return frameTime; },
     get frameCount() { return _frameCount; },
+    get frameStats(): DrawStats { return { ..._frameStats }; },
 
     makeBuffer(desc: BufferDesc): SgBuffer {
       const h = handle<SgBuffer>("SgBuffer");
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review — PR #36

**Verdict: APPROVED**

### Type-check
`npx tsc --noEmit` passes cleanly (workspace requires `--typeRoots node_modules/@types,node_modules` due to environment config; the PR source itself is error-free).

---

### Feature Evaluation

#### `BufferUsage.INDIRECT` + `gpuBufferUsage()`
Correct. `INDIRECT` maps to `GPUBufferUsage.COPY_DST | GPUBufferUsage.INDIRECT`. The default arm still covers `VERTEX | INDEX`, nothing broken.

#### `drawIndirect(indirectBuffer, indirectOffset?)`
Clean implementation. Dispatches `drawIndexedIndirect` vs `drawIndirect` based on `currentPipeline?.indexType !== IndexType.NONE`. Validates the buffer handle before use. Uses optional chaining correctly (no throw if `currentPipeline` is null — GPU validation will catch that at submission time, acceptable for the indirect path). Stats increment on `indirectDrawCalls`. No issues.

#### Bound buffer caching (`applyBindings`)
`boundVertexBuffers` is reset to `[]` at the top of `applyBindings` before re-populating — correct guard against stale slots from a prior binding. `boundIndexBuffer` is explicitly nulled in the `else` branch when no index buffer is provided. Both are also cleared on `beginPass`. Solid.

#### Draw validation
The auto-derive and validation logic uses `desc.size ?? desc.data?.byteLength ?? 0`. This is correct given that `makeBuffer` always stores the original `desc` on the slot — the GPU buffer's allocated size may be larger (padded), but the guard is conservative (validates against the descriptor intent, not the GPU allocation), which is the safe direction.

The `bytesPerIndex` computation is duplicated between the auto-derive block and the validation block — minor cosmetic redundancy, not a bug.

Validation throws before any GPU command is recorded. Error messages include the range and capacity, which is good for debugging.

#### Auto element count
When `numElements` is `undefined` or `0`, count derives from `bufferSize / bytesPerIndex - baseElement` (indexed) or `bufferSize / stride - baseElement` (non-indexed, guards `stride > 0`). Falls back to `0` if no buffers are bound. This mirrors the Sokol C convention and is straightforward to reason about.

#### `DrawStats` + `frameStats`
`_frameStats` is reset on `beginPass`, not `commit` — correct, stats should accumulate within a pass (multiple draw calls per pass are common). `frameStats` returns a shallow copy (`{ ..._frameStats }`) so callers can't mutate the live counter. `indirectDrawCalls` is tracked but `totalElements` is intentionally not accumulated for indirect draws (count is GPU-side at indirect dispatch time) — correct design choice.

---

### Minor Observations (non-blocking)
- The `bytesPerIndex` calculation could be extracted to a small helper to avoid the duplicated block, but this is a style preference, not a correctness issue.
- `drawIndirect` does not guard `!currentPipeline` explicitly (only `!passEncoder`), so calling it without a pipeline set will silently emit an indirect draw with an unset pipeline. This matches the GPU-validation-at-submission model, which is consistent with the rest of the codebase.

---

All four advertised features are correctly implemented, type-safe, and consistent with existing codebase conventions. Ready to merge.

Issue #18, opened 2026-04-19
Runtime pipeline rebuilding on shader change


rjwalters:
## Summary
Automatically rebuild render pipelines when their underlying shaders are recompiled, maintaining the same pipeline configuration.

## Requirements
- [ ] Track shader→pipeline dependencies
- [ ] On shader recompile, rebuild all dependent pipelines with same `PipelineDesc`
- [ ] Swap pipeline handles transparently (app code keeps working with same handle)
- [ ] Validate new pipeline before swapping (fall back to old on error)
- [ ] Async pipeline compilation (`createRenderPipelineAsync`) to avoid frame drops
- [ ] Event/callback on pipeline rebuild completion

## Design Notes
`PipelineSlot` stores the full `PipelineDesc` (`gfx.ts:38-41`), so rebuilding with updated shader modules is straightforward. The key challenge is making this transparent to user code.

## References
- Sokol: not supported
- WebGPU: `createRenderPipelineAsync()`

rjwalters:
## Curation Notes

### Dependencies

This issue **depends on #16** (Hot-reload support — dynamic shader recompilation). Issue #16 must land first because it introduces `recompileShader()` and establishes the updated `ShaderSlot` with new `GPUShaderModule` instances. Pipeline rebuilding fires in response to the events/callbacks that #16 emits on successful recompilation.

---

### Implementation Guidance

#### 1. Shader→Pipeline dependency tracking

Add a reverse-lookup map inside `createGfx` (`src/gfx.ts`, alongside the existing `pipelines` map at line 52):

```ts
// shader id → set of pipeline ids that reference it
const shaderPipelineDeps = new Map<number, Set<number>>();
```

In `makePipeline` (line 157), after `pipelines.set(h.id, ...)`, register the dependency:

```ts
const deps = shaderPipelineDeps.get(desc.shader.id) ?? new Set<number>();
deps.add(h.id);
shaderPipelineDeps.set(desc.shader.id, deps);
```

In `destroyPipeline` (line 241), remove the pipeline from its shader's dep set to avoid stale entries:

```ts
const slot = pipelines.get(pip.id);
if (slot) {
  const deps = shaderPipelineDeps.get(slot.desc.shader.id);
  deps?.delete(pip.id);
  pipelines.delete(pip.id);
}
```

#### 2. Async rebuild on shader recompile

[...]

PR #37, opened 2026-04-19
Add dynamic shader recompilation with error handling

### LINKED ISSUE #16: Hot-reload support — dynamic shader recompilation
## Summary
Recompile shaders at runtime without reloading the page, enabling rapid WGSL iteration.

## Requirements
- [ ] `recompileShader(handle, newSource)` API
- [ ] Shader compilation error reporting without crashing the app
- [ ] Vite HMR integration for `.wgsl` files
- [ ] WGSL file import support (vite plugin or raw import)
- [ ] Diff-based recompilation (skip if source unchanged)
- [ ] Callback/event on successful recompilation

## Design Notes
Shader source is already stored in `ShaderSlot` (`gfx.ts:30-34`), which enables recompilation. The challenge is invalidating and rebuilding dependent pipelines.

## References
- Sokol: not supported (C library, requires restart)
- Prior art: ShaderToy, three.js shader hot-reload

rjwalters:
Closes #16

## Summary

- Adds `ShaderRecompileResult` discriminated union type to `src/types.ts` (`{ ok: true; shader }` or `{ ok: false; vertexError?; fragmentError? }`)
- Extends the `Gfx` interface with `recompileShader(shd, sources, callback?): Promise<ShaderRecompileResult>`
- Implements the method in `src/gfx.ts` inside `createGfx`:
  - **Diff-based skip**: returns `{ ok: true }` immediately if both sources are unchanged, avoiding unnecessary GPU work
  - **Safe error capture**: uses `getCompilationInfo()` (not exceptions) to surface vertex/fragment compile errors; bad WGSL returns `{ ok: false }` without touching the slot or crashing the app
  - **Atomic commit**: only replaces `ShaderSlot` modules and source strings after both modules pass compilation info checks
  - **Optional callback**: invoked with the result for sync-style callers alongside the returned Promise
- Pipeline rebuilding for dependent `PipelineSlot` entries is intentionally out of scope — tracked in issue #18

## Test plan

- [ ] Diff guard: call `recompileShader` twice with identical source; second call skips `createShaderModule`
- [ ] Error isolation: pass broken WGSL; assert `result.ok === false`, error strings non-empty, app continues rendering
- [ ] Success path: valid new WGSL updates `ShaderSlot` modules and source strings
- [ ] Invalid handle: missing shader id returns `{ ok: false, vertexError: "Invalid shader handle" }`
- [ ] Callback: invoked correctly on both success and failure paths

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -1,7 +1,7 @@
 import {
   type SgBuffer, type SgImage, type SgSampler, type SgShader, type SgPipeline,
   type BufferDesc, type ImageDesc, type SamplerDesc, type ShaderDesc, type PipelineDesc,
-  type Bindings, type PassDesc, type Gfx, type DrawStats,
+  type Bindings, type PassDesc, type Gfx, type DrawStats, type ShaderRecompileResult,
   BufferUsage, IndexType, LoadAction, StoreAction, PixelFormat, PrimitiveType, CullMode, CompareFunc,
   FilterMode, WrapMode,
 } from "./types.js";
@@ -31,6 +31,8 @@ interface ShaderSlot {
   fragmentModule: GPUShaderModule;
   vertexEntry: string;
   fragmentEntry: string;
+  vertexSource: string;
+  fragmentSource: string;
 }
 
 interface PipelineSlot {
@@ -193,7 +195,9 @@ export function createGfx(
 
       const vertexEntry = desc.vertexEntry ?? "vs_main";
       const fragmentEntry = desc.fragmentEntry ?? "fs_main";
-      shaders.set(h.id, { vertexModule, fragmentModule, vertexEntry, fragmentEntry });
+      const vertexSource = combinedSource ?? desc.vertexSource!;
+      const fragmentSource = combinedSource ?? desc.fragmentSource ?? vertexSource;
+      shaders.set(h.id, { vertexModule, fragmentModule, vertexEntry, fragmentEntry, vertexSource, fragmentSource });
       return h;
     },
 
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review — PR #37: Add dynamic shader recompilation with error handling

**Verdict: APPROVE**

### Type Check
`tsc --noEmit` passes clean at commit `97c3c1a` with TypeScript 5.8.3. Zero errors.

### Criterion 1: Error handling without crashing
**PASS.** Two guard paths before any GPU work:
- Invalid handle → `{ ok: false, vertexError: "Invalid shader handle" }` returned and passed to callback; no throw.
- Compile errors captured via `getCompilationInfo()` (not exceptions) after `Promise.all` on both modules. Bad WGSL surfaces `vertexError`/`fragmentError` in the result union without touching the existing slot or crashing.

The `createShaderModule` call is intentionally non-throwing per WebGPU spec; errors are deferred to `getCompilationInfo()`, which is the correct pattern here.

### Criterion 2: Diff-based skip
**PASS.** After resolving `nextVertex`/`nextFragment` via nullish coalescing from the slot, a strict equality check (`=== `) short-circuits before any `createShaderModule` call, returning `{ ok: true, shader: shd }` immediately. This prevents unnecessary GPU object creation on redundant calls.

### Criterion 3: Atomic slot update
**PASS.** Both `vertInfo` and `fragInfo` are awaited together with `Promise.all`. Only after both compilation info checks pass (no `"error"` messages in either) are the slot fields mutated — `vertexModule`, `fragmentModule`, `vertexSource`, `fragmentSource` — all in one synchronous block. There is no partial state where one module is updated and the other is not.

### Callback / Promise dual API
The optional `callback` is invoked with the same result object on every exit path (invalid handle, compile error, diff skip, success) before `return`. The Promise resolves with the same value. Consistent and correct.

- The `{ ok: true }` diff-skip result correctly returns `shader: shd` (the input handle), consistent with the success path — good.
- Pipeline rebuilding for dependent `PipelineSlot` entries is explicitly deferred to issue #18 with a comment; the current slot mutation (updating modules) is complete enough for callers that manually rebuild pipelines.
- `ShaderRecompileResult` is exported via `export * from "./types.js"` in `src/index.ts` — no extra wiring needed.

### Summary
All three evaluation criteria are fully satisfied. The implementation is clean, type-safe, and follows the no-crash contract required for live-reload tooling. Ready to merge.

PR #39, opened 2026-04-19
Add uniform buffer overflow detection and ring-buffered allocation

### LINKED ISSUE #8: Uniform binding with dynamic offsets
## Summary
Bind uniform data to shaders using a shared uniform buffer with dynamic offsets, aligned to WebGPU requirements.

## Current State
`gfx.ts:applyUniforms` writes uniform data to a 64KB staging buffer with 256-byte alignment. Creates a bind group per draw with dynamic offset.

## Requirements
- [x] Shared uniform buffer with dynamic offsets (`gfx.ts:305-319`)
- [x] 256-byte alignment for WebGPU (`gfx.ts:310`)
- [x] Per-frame offset reset (`gfx.ts:280`)
- [ ] Multiple bind groups (currently only group 0)
- [ ] Uniform buffer overflow detection
- [ ] Ring-buffered uniform allocation for multi-frame in-flight
- [ ] Typed uniform helpers (e.g., `applyUniformsFromObject`)
- [ ] Reduce bind group creation overhead (cache/reuse)

## References
- Sokol: `sg_apply_uniforms()`
- WebGPU: `GPUQueue.writeBuffer()`, `setBindGroup()` with dynamic offsets

rjwalters:
## Summary

- **Overflow detection**: `applyUniforms` now checks `alignedOffset + slotSize > UNIFORM_BUFFER_SIZE` before calling `writeBuffer` and throws a descriptive error, preventing silent GPU validation errors when a pass issues too many uniform updates.
- **Ring-buffered allocation**: `createGfx` allocates `NUM_FRAMES_IN_FLIGHT` (2) `GPUBuffer`s up front. `commit()` advances `uniformFrameIndex` modulo `NUM_FRAMES_IN_FLIGHT`, ensuring the CPU never writes into a buffer the GPU is still reading from a prior frame.
- **Reduced bind group overhead**: One `GPUBindGroup` is pre-created per ring slot at init time. `applyUniforms` now calls `passEncoder.setBindGroup` with the cached group and a dynamic offset — eliminating the `device.createBindGroup` allocation that previously occurred on every draw call. `makePipeline` also reuses the shared `uniformBindGroupLayout` instead of creating a new layout per pipeline.

## Test plan

- [ ] Overflow detection — call `applyUniforms` in a loop with 256-byte payloads until 64 KB is exhausted; assert the 257th call throws with a clear message.
- [ ] Ring-buffer correctness — render two frames back-to-back with different uniform values; assert frame N+1 does not corrupt frame N's result (GPU readback via `copyTextureToBuffer`).
- [ ] Bind group cache hit rate — instrument `device.createBindGroup` with a spy; render 100 frames with a single `applyUniforms` call each; assert `createBindGroup` is called at most `NUM_FRAMES_IN_FLIGHT` times total.
- [ ] Regression — run existing triangle/quad samples and assert uniform data is delivered correctly after all changes.

Closes #8

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -151,27 +151,51 @@ export function createGfx(
   let lastFrameTime = 0;
   let _frameCount = 0;
   let uniformOffset = 0;
-  let uniformBuffer: GPUBuffer | null = null;
-  let uniformBindGroup: GPUBindGroup | null = null;
   let boundVertexBuffers: BufferSlot[] = [];
   let boundIndexBuffer: BufferSlot | null = null;
   let _frameStats: DrawStats = { drawCalls: 0, totalElements: 0, indirectDrawCalls: 0 };
 
-  const UNIFORM_BUFFER_SIZE = 65536; // 64KB uniform staging
+  const UNIFORM_BUFFER_SIZE = 65536; // 64KB uniform staging per ring slot
   if (UNIFORM_BUFFER_SIZE > device.limits.maxUniformBufferBindingSize) {
     throw new Error(
       `UNIFORM_BUFFER_SIZE (${UNIFORM_BUFFER_SIZE}) exceeds device limit maxUniformBufferBindingSize (${device.limits.maxUniformBufferBindingSize})`
     );
   }
+  const NUM_FRAMES_IN_FLIGHT = 2;
+  let uniformFrameIndex = 0;
 
-  function ensureUniformBuffer() {
-    if (!uniformBuffer) {
-      uniformBuffer = device.createBuffer({
-        size: UNIFORM_BUFFER_SIZE,
-        usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
-      });
-    }
-    return uniformBuffer;
+  // Shared bind group layout for the uniform ring buffers (group 0, binding 0, dynamic offset)
+  const uniformBindGroupLayout = device.createBindGroupLayout({
+    entries: [{
+      binding: 0,
+      visibility: GPUShaderStage.VERTEX | GPUShaderStage.FRAGMENT,
+      buffer: { type: "uniform", hasDynamicOffset: true },
+    }],
+  });
+
+  // Allocate NUM_FRAMES_IN_FLIGHT uniform buffers and one bind group per slot up front
+  const uniformBuffers: GPUBuffer[] = [];
+  const uniformBindGroups: GPUBindGroup[] = [];
+  for (let i = 0; i < NUM_FRAMES_IN_FLIGHT; i++) {
+    const buf = device.createBuffer({
+      size: UNIFORM_BUFFER_SIZE,
+      usage: GPUBufferUsage.UNIFORM | GPUBufferUsage.COPY_DST,
+      label: `uniform_ring_${i}`,
+    });
+    uniformBuffers.push(buf);
+    uniformBindGroups.push(device.createBindGroup({
+      layout: uniformBindGroupLayout,
+      entries: [{ binding: 0, resource: { buffer: buf, size: 256 } }],
+      label: `uniform_bind_group_${i}`,
+    }));
+  }
+
+  function currentUniformBuffer() {
+    return uniformBuffers[uniformFrameIndex];
+  }
+
+  function currentUniformBindGroup() {
+    return uniformBindGroups[uniformFrameIndex];
   }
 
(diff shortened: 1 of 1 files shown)


rjwalters:
## Loom Review — PR #39: Uniform Ring Buffer

**Type-check**: `npx tsc --noEmit` passes cleanly (0 errors).

---

### Evaluation

#### Overflow detection
The guard `alignedOffset + slotSize > UNIFORM_BUFFER_SIZE` in `applyUniforms` is placed correctly **before** `writeBuffer`, and the error message is descriptive. The pre-check uses `slotSize = Math.max(data.byteLength, 256)`, which correctly accounts for the 256-byte alignment slot consumed even by small payloads.

#### Ring buffer rotation
`uniformFrameIndex` advances in `commit()` via `(uniformFrameIndex + 1) % NUM_FRAMES_IN_FLIGHT`. `uniformOffset` resets in `beginPass()`, not `commit()` — which is correct; a frame may contain multiple passes, each sharing the same ring slot. Ring index only flips when the frame is submitted.

#### Pre-created bind groups
Two bind groups are created at init time with no `createBindGroup` in the draw hot path.

**Bug found — bind group size is hardcoded to 256 bytes.**

```ts
uniformBindGroups.push(device.createBindGroup({
  layout: uniformBindGroupLayout,
  entries: [{ binding: 0, resource: { buffer: buf, size: 256 } }],
```

The old code used `size: Math.max(data.byteLength, 256)` per upload, which correctly sized the binding window to the actual struct. Pre-creating the bind group fixes the size at 256. WebGPU binds using `[offset, offset + bindingSize)` from the buffer; if the shader reads a uniform struct larger than 256 bytes (e.g. a 512-byte transform block), the GPU sees a 256-byte window and the upper half is out-of-range, producing a validation error or silent corruption.

**The correct fix** is to use `size: UNIFORM_BUFFER_SIZE` (bind the full buffer and let the dynamic offset restrict the start), or to defer `createBindGroup` until the first `applyUniforms` of each pipeline size class (a bind group per buffer × size pair). Using `size: UNIFORM_BUFFER_SIZE` is the simplest path and is valid because WebGPU allows binding the whole buffer when `minBindingSize` is 0 (the default).

#### Shared layout
`uniformBindGroupLayout` is created once at `createGfx` time and reused in every `makePipeline` call, eliminating the per-pipeline `createBindGroupLayout` allocation. Correct.

---

### Minor observations
- `uniformOffset` is no longer reset in `commit()` — only in `beginPass()`. If `commit()` fires without a preceding `beginPass()` on that frame the stale offset carries into the next frame's first pass. Unlikely in practice but worth a comment.
- `NUM_FRAMES_IN_FLIGHT = 2` is not exported or part of the public `Gfx` type, so callers cannot pipeline more aggressively. Acceptable for now; doc comment noting the constant would help future readers.

---

### Verdict
**Changes requested** — the fixed bind group `size: 256` is a correctness bug for any uniform struct larger than 256 bytes.

rjwalters:
## Loom Judge Review -- APPROVED

All three fixes from the previous rejection are correctly addressed:

1. **Bind group size**: Now uses `UNIFORM_BUFFER_SIZE` (65536) instead of hardcoded 256 for the bind group entry `size`. With `hasDynamicOffset: true`, this correctly represents the maximum binding range within the buffer, allowing uniform structs of any size up to the buffer limit.

2. **uniformOffset reset**: Properly reset to 0 in `commit()`, so each frame starts fresh in its own ring slot. Previously this was only reset in `beginPass()`, meaning multi-pass frames within a single commit would accumulate offsets across ring slot boundaries.

3. **NUM_FRAMES_IN_FLIGHT export**: Exported at module scope from `gfx.ts`.

### Other observations

- Ring buffer alternation logic is correct: `uniformFrameIndex` advances modulo `NUM_FRAMES_IN_FLIGHT` in `commit()`.
- Overflow detection is correct: checks before `writeBuffer`, not after.
- Shared `uniformBindGroupLayout` eliminates per-pipeline layout redundancy.
- Pre-allocated bind groups eliminate per-draw-call `createBindGroup` overhead.

**Nit** (non-blocking): `NUM_FRAMES_IN_FLIGHT` is exported from `gfx.ts` but not re-exported from `index.ts`. Consumers using the package entry point won't see it. Consider adding it to the named exports in `index.ts` if it's meant to be part of the public API.

Type check passes cleanly.

rjwalters:
Merge conflicts resolved via rebase onto latest main.

PR #40, opened 2026-04-19
Add WebAudio wrapper module (sokol_audio equivalent)

### LINKED ISSUE #19: sokol_audio equivalent — WebAudio API wrapper
## Summary
Provide a simple audio playback API wrapping the Web Audio API, matching the design philosophy of sokol_audio.

## Requirements
- [ ] `AudioDesc` configuration (sample rate, channels, buffer size)
- [ ] Callback-based audio generation (fill buffer each quantum)
- [ ] `initAudio(desc)` / `shutdownAudio()` lifecycle
- [ ] AudioWorklet-based processing (low latency)
- [ ] Volume control
- [ ] Suspend/resume handling (browsers require user gesture to start audio)
- [ ] Multiple audio streams
- [ ] Simple sound effect playback helper

## Design Notes
sokol_audio uses a push model with a callback that fills a float buffer. Web Audio API's `AudioWorklet` provides the equivalent low-latency callback mechanism. Fall back to `ScriptProcessorNode` for broader compatibility.

## References
- Sokol: `saudio_setup()`, `saudio_push()`, `saudio_sample_rate()`
- Web: `AudioContext`, `AudioWorklet`, `AudioWorkletProcessor`

rjwalters:
Closes #19

## Summary

- Adds `createAudio(desc: AudioDesc): Promise<Audio>` factory in `src/audio.ts` using AudioWorklet via an inline Blob URL (no external asset files required)
- Adds `AudioDesc`, `Audio`, `SaudioStream`, and `AudioCallback` types to `src/types.ts`
- Exports `createAudio` from `src/index.ts`

## Design notes

- **AudioWorklet** (not the deprecated `ScriptProcessorNode`) runs the audio pipeline in a dedicated rendering thread; the fill callback is invoked on the main thread in response to worklet `'request'` messages, then per-channel buffers are posted back
- **GainNode** sits between the worklet and `context.destination`; `setVolume(v)` sets `gainNode.gain.value` directly
- **Suspend/resume**: `context.resume()` is called eagerly at construction; `Audio.resume()` is async and throws after `shutdown()` so callers can surface a "tap to start" prompt while `!audio.isRunning`
- **Visibility handling**: a `visibilitychange` listener suspends when the tab hides and resumes when it returns (unless the user explicitly called `suspend()`)
- **`SaudioStream`** handle type is defined in `types.ts` for future multi-stream support; initial implementation uses one stream per `Audio` instance

## Test plan

- [ ] `createAudio({ streamCallback })` constructs without throwing in a browser dev environment
- [ ] `audio.isRunning` is `false` before a user gesture; becomes `true` after `audio.resume()` inside a click handler
- [ ] Writing a 440 Hz sine wave in `streamCallback` produces an audible tone
- [ ] `audio.setVolume(0)` silences; `setVolume(1)` restores without discontinuity
- [ ] `audio.suspend()` / `resume()` pause and restart callback delivery
- [ ] `audio.shutdown()` closes the `AudioContext`; calling `resume()` afterwards throws
- [ ] Switching tabs suspends audio; restoring the tab resumes it
- [ ] `npx tsc --noEmit` passes with zero errors (verified)

### DIFF
--- a/src/audio.ts
+++ b/src/audio.ts
@@ -1,0 +1,157 @@
+import { type AudioDesc, type Audio } from "./types.js";
+
+// AudioWorklet processor source — runs in the audio rendering thread.
+// The main thread sends filled channel buffers back in response to 'request' messages.
+const PROCESSOR_SOURCE = `
+class SokolProcessor extends AudioWorkletProcessor {
+  constructor(options) {
+    super();
+    this._numChannels = options.processorOptions.numChannels;
+    this._bufferFrames = options.processorOptions.bufferFrames;
+    this._pending = null;
+    this.port.onmessage = (e) => {
+      if (e.data.type === 'buffer') {
+        this._pending = e.data.channels;
+      }
+    };
+    // Request the first buffer immediately
+    this.port.postMessage({ type: 'request', numFrames: this._bufferFrames });
+  }
+  process(_inputs, outputs) {
+    const out = outputs[0];
+    if (this._pending) {
+      for (let ch = 0; ch < out.length && ch < this._pending.length; ch++) {
+        out[ch].set(this._pending[ch]);
+      }
+      this._pending = null;
+    }
+    // Request the next buffer
+    this.port.postMessage({ type: 'request', numFrames: this._bufferFrames });
+    return true;
+  }
+}
+registerProcessor('sokol-processor', SokolProcessor);
+`;
+
+export async function createAudio(desc: AudioDesc): Promise<Audio> {
+  const numChannels = desc.numChannels ?? 2;
+  const bufferFrames = desc.bufferFrames ?? 128;
+  const volume = desc.volume ?? 1.0;
+
+  // Create AudioContext with requested sample rate (or browser default)
+  const contextOptions: AudioContextOptions = {};
+  if (desc.sampleRate !== undefined) {
+    contextOptions.sampleRate = desc.sampleRate;
+  }
+  const context = new AudioContext(contextOptions);
+
+  // Install AudioWorklet processor via inline Blob URL (keeps lib self-contained)
+  const blob = new Blob([PROCESSOR_SOURCE], { type: "application/javascript" });
+  const blobUrl = URL.createObjectURL(blob);
+  try {
+    await context.audioWorklet.addModule(blobUrl);
+  } finally {
+    URL.revokeObjectURL(blobUrl);
+  }
+
+  // Create the worklet node
+  const workletNode = new AudioWorkletNode(context, "sokol-processor", {
+    numberOfInputs: 0,
+    numberOfOutputs: 1,
+    outputChannelCount: [numChannels],
+    processorOptions: { numChannels, bufferFrames },
+  });
+
+  // Create gain node for volume control, connect graph: worklet -> gain -> destination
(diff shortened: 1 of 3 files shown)


rjwalters:
## Loom Judge Review — WebAudio Wrapper

**Verdict: APPROVE**

`npx tsc --noEmit` passes with zero errors/warnings.

---

### AudioWorklet usage
Correct and idiomatic. Processor is embedded as a string literal and loaded via an inline Blob URL (created, used, then immediately revoked in a `try/finally` — no leak). `registerProcessor` is called at module scope. Using `AudioWorkletNode` over the deprecated `ScriptProcessorNode` is the right call.

One minor note: the worklet `process()` requests the next buffer unconditionally on every render quantum, even when `_pending` is null and silence was output. This means the main thread is always one request behind, which is acceptable at 128-frame quanta but could cause a double-request on the very first quantum. In practice, the constructor already fires the first request, so `process()` fires a second request immediately — this results in two in-flight requests at startup. Not a correctness bug (the worklet just buffers one extra frame), but worth noting.

### Callback pattern
Clean. `streamCallback(buffer, numFrames, numChannels)` matches the `sokol_audio` interleaved convention. The scratch buffer (`interleavedBuf`) is allocated once and reused across callbacks — no per-call allocation on the hot path (the per-channel de-interleave arrays are allocated per message, but that is on the main thread, not the audio thread). The interleaved → planar split is correct.

### Volume control
`gainNode.gain.value = v` is a direct, sample-accurate assignment appropriate for non-automated volume changes. No range clamping (e.g. to `[0, 1]`) — callers can pass values outside that range, which the Web Audio spec allows (gain > 1 amplifies, negative inverts phase). This is acceptable given the library's low-level nature, but could be a footgun for users expecting sokol_audio's `saudio_set_volume` semantics.

### Suspend / resume
- `userSuspended` flag correctly prevents the `visibilitychange` handler from auto-resuming after an explicit `suspend()` call. 
- `resume()` correctly throws after `shutdown()` with a clear error message.
- `suspend()` silently returns after `shutdown()` (idempotent) — consistent with how `shutdown()` itself is idempotent.
- Eagerly calling `context.resume()` at construction is fine; the promise rejection is silently swallowed, which is correct (autoplay policy may block it until a user gesture).

### Cleanup
All resources are properly released in `shutdown()`: event listener removed, both nodes disconnected, `AudioContext` closed. The `_shutdown` guard prevents double-shutdown. No orphaned intervals, timeouts, or object URLs.

### Types
`AudioCallback`, `AudioDesc`, `Audio`, and `SaudioStream` are all clean and well-documented. `SaudioStream` is correctly forward-declared but unused in the implementation (noted in a comment). No unused types that would trip `noUnusedLocals`.

### Minor issues (non-blocking)
1. **Double in-flight request at startup**: constructor posts one request, then `process()` immediately posts another before the first reply arrives. Low priority.
2. **No volume clamping**: `setVolume` accepts any `number`. Consider documenting the expected range.
3. **`debugText.ts`** appears in `src/` but is not referenced by `index.ts` — pre-existing, not introduced by this PR.

---

Overall this is a solid, production-quality module. The design is faithful to `sokol_audio`, the cleanup story is complete, and the type-check is clean.

PR #43, opened 2026-04-19
Add key repeat, focus/blur, pointer lock, drag-drop, and gamepad input

### LINKED ISSUE #13: Input event system (keyboard, mouse, touch, resize)
## Summary
Capture and dispatch keyboard, mouse, touch, and resize events through a unified event callback.

## Current State
`app.ts` registers listeners for mouse, keyboard, touch, and resize events. Events are dispatched through a single `event` callback on `AppDesc`.

## Requirements
- [x] Mouse events: down, up, move, wheel (`app.ts:58-61`)
- [x] Keyboard events: keydown, keyup (`app.ts:62-63`)
- [x] Touch events: start, move, end (`app.ts:65-75`)
- [x] Resize via ResizeObserver (`app.ts:77-81`)
- [ ] Pointer lock support (`requestPointerLock`)
- [ ] Gamepad input
- [ ] Focus/blur events
- [ ] Drag and drop
- [ ] Key repeat filtering (distinguish held keys from new presses)
- [ ] Normalized coordinates option (0-1 range)
- [ ] Event queue mode (buffer events, process in frame)

## References
- Sokol: `sapp_event`, `SAPP_EVENTTYPE_*`
- Web: DOM events, Pointer Events API

rjwalters:
Closes #13

### src/types.ts
- `AppEvent.keyRepeat?: boolean` — populated from `KeyboardEvent.repeat`
- `AppEvent.mouseNormX/mouseNormY?: number` — opt-in normalized mouse coords (0–1)
- `AppEvent.touches` extended with optional `normX/normY` per touch point
- `AppEvent.gamepadIndex/gamepadButton/gamepadAxis/gamepadValue` — gamepad event fields
- `AppEvent.files?: FileList` — for drag-and-drop file access
- `AppDesc.normalizedCoords?: boolean` — enables normalized coordinate fields
- `AppDesc.pointerLock?: boolean` — enables pointer lock on mousedown
- `AppDesc.eventQueue?: boolean` — buffers events and flushes at frame start
- New `AppEventType` values: `FOCUS`, `BLUR`, `POINTER_LOCK`, `POINTER_UNLOCK`, `GAMEPAD_DOWN`, `GAMEPAD_UP`, `GAMEPAD_AXIS`, `GAMEPAD_CONNECTED`, `GAMEPAD_DISCONNECTED`, `DRAG_OVER`, `DRAG_LEAVE`, `DROP`

### src/app.ts
- `keydown` listener now sets `keyRepeat: e.repeat`
- `focus`/`blur` listeners on `window`
- Pointer lock: `canvas.requestPointerLock()` on `mousedown` when `desc.pointerLock`; `pointerlockchange`/`pointerlockerror` listeners on `document`
- Drag-and-drop: `dragover` (with `preventDefault`), `dragleave`, `drop` listeners on canvas
- Gamepad: `gamepadconnected`/`gamepaddisconnected` listeners on `window`; per-frame `pollGamepads()` diffs button pressed state and axis values against previous snapshot
- Event queue: `dispatch()` buffers when `desc.eventQueue`; `flushEvents()` drains at start of each frame
- `listen()` helper updated with overload signature to accept non-HTMLElementEventMap event types (document/window-specific events)

## Test Plan
- [ ] Hold a key >500ms: first event has `keyRepeat: false`, subsequent have `keyRepeat: true`
- [ ] Click away from tab: `BLUR` fires. Click back: `FOCUS` fires
- [ ] `normalizedCoords: true`: mouse at corners produces `{0,0}`, `{1,0}`, `{0,1}`, `{1,1}`
- [ ] `pointerLock: true`: click canvas → `POINTER_LOCK` fires; Escape → `POINTER_UNLOCK` fires
- [ ] `eventQueue: true`: multiple events before next frame arrive in order at frame start
- [ ] Gamepad connect/disconnect → `GAMEPAD_CONNECTED`/`GAMEPAD_DISCONNECTED`; button/axis changes → `GAMEPAD_DOWN`/`GAMEPAD_UP`/`GAMEPAD_AXIS`
- [ ] Drag file onto canvas: `DRAG_OVER` during hover, `DROP` with populated `files`, `DRAG_LEAVE` on exit
- [ ] Cleanup function removes all new listeners


rjwalters:
## Loom Judge Review — PR #43 (input events expansion)

**TypeScript:** `npx tsc --noEmit` passes with zero errors.

---

### Issues requiring changes

#### 1. `POINTER_UNLOCK` enum value is semantically wrong (blocking)

```ts
// types.ts
POINTER_LOCK   = "pointerlockchange",
POINTER_UNLOCK = "pointerlockerror",   // BUG
```

`POINTER_UNLOCK` is assigned the string `"pointerlockerror"`, but the actual browser event that fires when the pointer is *unlocked* (e.g. user presses Escape) is `"pointerlockchange"` — the same event as lock. The error event (`pointerlockerror`) means lock acquisition *failed*, not that the pointer was released. A consumer doing `if (ev.type === AppEventType.POINTER_UNLOCK)` will never match a normal unlock because the dispatched event carries `AppEventType.POINTER_LOCK` (= `"pointerlockchange"`).

The enum values for synthetic/virtual events (like the gamepad ones: `"gamepad_down"`, `"gamepad_axis"`, etc.) are already correctly using opaque strings. Fix:

```ts
POINTER_LOCK   = "pointer_lock",
POINTER_UNLOCK = "pointer_unlock",
```

And update the dispatch site in `app.ts` accordingly (it does not pass the string to the DOM, so there's no other risk).

---

#### 2. Touch normalized coords divide already-scaled pixels by `canvas.width` (off by `pixelRatio²`)

```ts
// app.ts
const x = t.clientX * pixelRatio;   // pixel-ratio-scaled physical pixels
...
normX: normX(x),                     // x / canvas.width
```

`normX` divides by `canvas.width`, which is also in physical pixels (set via `resize()`). So `x / canvas.width` = `(clientX * pixelRatio) / (clientWidth * pixelRatio)` = `clientX / clientWidth` — this is actually *correct* because `pixelRatio` cancels. However, the touch origin is `clientX`, not `offsetX`, so the normalization doesn't account for the canvas position in the page. For a full-page canvas this is fine, but for an inset canvas, `clientX / canvas.width` can exceed [0,1]. The mouse events use `offsetX` (canvas-relative), which is correct. Touch should use `clientX - canvas.getBoundingClientRect().left` (or equivalent) before normalizing. This is a pre-existing issue for touch *positions*, but the PR is the right moment to fix it since it's adding the normalization path.

---

#### 3. Drag-and-drop listeners are always registered regardless of any opt-in flag

Pointer lock listeners are gated on `desc.pointerLock`. Drag-drop listeners are unconditional, meaning every app receives `dragover`/`dragleave`/`drop` events and the canvas prevents the browser's default drag behavior (file open, etc.) on every drop. There should be a `desc.dragDrop?: boolean` opt-in flag mirroring `pointerLock`, or at minimum the `dragover preventDefault()` should be removed when there is no intent to handle drops.

---

#### 4. `pollGamepads()` fires even when no gamepad is connected (minor / perf)

`pollGamepads()` is called unconditionally every frame. `navigator.getGamepads()` is a synchronous API with up to 4 slots and is cheap, but it should still only run when at least one gamepad is connected. Track a connected count with the `gamepadconnected`/`gamepaddisconnected` listeners and skip the poll when the count is zero.

---

### Non-blocking observations (no changes required)

- **`void e;` in `dragleave`**: suppresses the unused-variable lint warning but is a style smell; just use `(_e)` or omit the parameter entirely since it's never read.
- **`keyup` missing `keyRepeat`**: `KeyboardEvent.repeat` is always `false` on keyup, so it's intentionally omitted — fine.
- **Event queue `splice(0)` idiom**: correct and safe; drains atomically before the frame callback.
- **`prevGamepadButtons` initialized as `[]` on first poll**: `undefined ?? []` means the first poll always sees every currently-pressed button as a transition from `false` → `true`. That's acceptable (it mirrors a connect event), but worth a comment.
- **`listen()` overload**: the dual overload approach is clean; cleanup via `eventMap` correctly covers all new listeners.

---

### Summary

Three blocking issues:
1. `POINTER_UNLOCK` carries the wrong string value — switch to an opaque string like the gamepad events.
2. Touch `clientX`-based normalization is canvas-position-unaware — use canvas-relative coordinates.
3. Drag-drop listeners should be opt-in (add `desc.dragDrop?: boolean` guard).

Please address these and re-request review.

rjwalters (commented):
## Loom Judge Review: Approved

All three previously requested fixes are correctly implemented:

1. **POINTER_UNLOCK enum value** -- Changed from `"pointerlockerror"` to `"pointer_unlock"`, a distinct synthetic value separate from the browser's `pointerlockchange` event. Correct.

2. **Touch normalization** -- `touchPos()` now uses `canvas.getBoundingClientRect()` to compute canvas-relative coordinates (`clientX - rect.left`) before applying `pixelRatio`. This matches how mouse events use `offsetX`. Correct.

3. **Drag-drop opt-in** -- All drag/drop listeners are gated behind `if (desc.dragDrop)`, so apps that don't handle files retain native drag behavior. Correct.

**Minor observations (non-blocking):**
- Gamepad axis diffing uses exact float equality, which will fire many events from stick jitter. Consider adding an optional deadzone threshold in a future iteration.
- `pollGamepads()` runs every frame even with no gamepads connected. `navigator.getGamepads()` is cheap so this is fine for now.

Type check: no errors in `src/app.ts` or `src/types.ts`. (Pre-existing `gfx.ts` errors are from a separate branch and unrelated to this PR.)

LGTM.

PR #45, opened 2026-04-19
Add async asset loading module (sokol_fetch equivalent)

### LINKED ISSUE #20: sokol_fetch equivalent — async asset loading
## Summary
Async asset loading with progress tracking, cancellation, and caching — equivalent to sokol_fetch.

## Requirements
- [ ] `fetch(url, callback)` with typed responses (ArrayBuffer, string, JSON, Image)
- [ ] Progress reporting (bytes loaded / total)
- [ ] Request cancellation
- [ ] Concurrent request limiting (configurable max in-flight)
- [ ] Request prioritization
- [ ] In-memory caching with configurable eviction
- [ ] Batch loading with completion callback
- [ ] Integration with `makeImage` (load URL → create texture)
- [ ] Integration with `makeShader` (load WGSL file → create shader)

## Design Notes
sokol_fetch manages a channel/lane system for concurrent requests. The browser's `fetch()` API handles most of this natively, so the wrapper should focus on ergonomics (typed results, auto GPU resource creation) rather than reimplementing concurrency control.

## References
- Sokol: `sfetch_setup()`, `sfetch_send()`, `sfetch_handle_t`
- Web: `fetch()`, `Response`, `AbortController`

rjwalters:
## Summary

- Adds `createSfetch(desc?)` factory in `src/fetch.ts` — the sokol_fetch equivalent for browser async asset loading
- Implements typed fetch responses: `arraybuffer`, `text`, `json`, and `image` (via `createImageBitmap` for off-thread decode)
- Progress reporting via `ReadableStream` chunked reads with `onProgress(FetchProgress)` callback
- Per-request and global cancellation via `AbortController` (composes caller signal + internal signal + root signal)
- Concurrent request limiting with a priority queue (`maxConcurrent`, default 6)
- LRU in-memory cache with configurable `cacheCapacity` (0 = disabled)
- Batch loading with `batch(requests, onAllDone)` counter pattern
- `fetchImage(gfx, req)` integration helper: fetches image → `createImageBitmap` → `gfx.makeImage` → `gfx.writeImageBitmap`
- `fetchShader(gfx, req)` integration helper: fetches WGSL text → `gfx.makeShader`
- Adds `writeImageBitmap(img, bitmap)` to the `Gfx` interface and `createGfx` implementation
- Exports `createSfetch` and all new types from `src/index.ts`

## Test plan

- [ ] Basic fetch — ArrayBuffer: assert `result instanceof ArrayBuffer`
- [ ] Basic fetch — text: assert `typeof result === "string"`
- [ ] Basic fetch — JSON: assert parsed object equality
- [ ] Basic fetch — image: assert `result instanceof ImageBitmap` with correct dimensions
- [ ] Progress reporting: mock slow stream; assert monotonically increasing `loaded`
- [ ] AbortController cancellation: abort mid-flight; assert `onError` fires and queue advances
- [ ] `cancelAll()`: queue 3 requests, cancelAll before completion; assert all `onError` fire
- [ ] Concurrency limit: `maxConcurrent: 2`, enqueue 5; assert max 2 in-flight
- [ ] Priority ordering: priorities 0, 5, 1; assert dequeue order 5 → 1 → 0
- [ ] Cache hit: fetch same URL twice; assert `globalThis.fetch` called once
- [ ] Cache eviction: `cacheCapacity: 2`, fetch 3 URLs; first evicted on 4th call
- [ ] `clearCache()`: fetch, clear, fetch again; assert fetched twice
- [ ] Batch `onAllDone`: assert fires exactly once after all `onDone`s
- [ ] `fetchImage` → `SgImage`: valid handle with correct texture dimensions
- [ ] `fetchShader` → `SgShader`: valid handle
- [ ] Error handling: 404 URL; assert `onError` fires, `onDone` does not

Closes #20


rjwalters:
## Loom Judge Review — PR #45: async asset loading

**TypeScript:** `tsc --noEmit` passes clean on all new PR code. (Pre-existing error in `gfx.ts` line 216 is on `main`; not introduced here.)

---

### Criteria Evaluation

#### Typed responses — PASS
All four response types (`arraybuffer`, `text`, `json`, `image`) are correctly dispatched. The `FetchRequest<T>` generic flows correctly from the public API through the internal cast to `FetchRequest<unknown>` and back to `T` in `onDone`. The image fast-path (blob → `createImageBitmap` without buffering) is a solid optimization.

#### Progress reporting — PASS
`FetchProgress` (`loaded`, `total`, `ratio`) is emitted on every chunk. Correctly handles servers that omit `Content-Length` by reporting `total: 0, ratio: 0`. Progress is skipped on the image fast-path (no-progress image request), which is consistent with the design.

#### Cancellation — PASS
`anySignal` correctly composes three signals (internal per-request, root, caller). `cancelAll()` aborts `rootController` (hitting in-flight fetches) and drains the pending queue firing `onError` for each entry. The rotation of `rootController` to a new instance before draining is correct — subsequent requests will use the fresh controller.

#### Concurrency limiting — PASS
`pump()` correctly enforces `maxConcurrent`. The `finally` → `pump()` chain ensures the slot is released and the next queued entry starts promptly. Priority sort on each `pump()` call is correct, though it's O(n log n) per dequeue event; a heap would scale better if queues ever grow large. Not a blocking concern for typical asset counts.

#### LRU cache — PASS
Map insertion-order trick correctly implements LRU eviction. `capacity === 0` guard prevents any caching. `clearCache()` works. Cache is keyed on raw URL strings — callers with query-string variants should be aware that `?v=1` and `?v=2` are independent keys, which is the expected behavior.

---

### Bugs — Changes Required

#### BUG 1 (blocking): `batch()` never fires `onAllDone` when any request errors

`batch()` wraps `req.onDone` to decrement `remaining` but does **not** wrap `req.onError`. If any request in the batch fails, its `onDone` is never called, `remaining` never reaches 0, and `onAllDone` is permanently suppressed.

```ts
// fetch.ts line 262-269 — current
for (const req of requests) {
  const originalDone = req.onDone;
  req.onDone = (result, url) => {
    originalDone(result, url);
    if (--remaining === 0) onAllDone();
  };
  ctx.fetch(req);
}
```

Fix: also wrap `onError` to decrement `remaining`:

```ts
for (const req of requests) {
  const originalDone = req.onDone;
  const originalError = req.onError;
  req.onDone = (result, url) => {
    originalDone(result, url);
    if (--remaining === 0) onAllDone();
  };
  req.onError = (err, url) => {
    originalError?.(err, url);
    if (--remaining === 0) onAllDone();
  };
  ctx.fetch(req);
}
```

#### BUG 2 (blocking): `batch()` mutates the caller's request objects

`batch()` overwrites `req.onDone` (and currently skips `req.onError`) directly on the objects passed in by the caller. If a caller reuses request objects or inspects them after calling `batch()`, they will see the wrapped callbacks — a surprising mutation of input. The method should either deep-clone each request or document that `batch()` takes ownership and the caller must not reuse the objects. Prefer cloning:

```ts
const wrapped = { ...req, onDone: ..., onError: ... };
ctx.fetch(wrapped);
```

---

### Design Notes (non-blocking)

**`fetchShader` passes the same source string to both `vertexSource` and `fragmentSource`.** This works for combined-shader formats but will be incorrect for the typical case where vertex and fragment shaders are in separate files. Consider either:
- Accepting `{ vertexUrl, fragmentUrl }` and doing two fetches, or
- Documenting clearly that `fetchShader` is only for single-file combined WGSL and the caller is responsible for the two-URL case.

As-is, this is a footgun waiting to be triggered.

---

### Summary

Two blocking bugs in `batch()`: errors silently suppress `onAllDone`, and caller request objects are mutated. Both are straightforward fixes. The rest of the implementation is solid — typed generics, LRU cache, priority queue, cancellation, and progress tracking are all correct. Please fix and re-request review.

rjwalters (commented):
## Judge Review: Approved

**Both previously-requested fixes verified:**

1. **batch() onAllDone on error**: The `onError` wrapper now correctly decrements `remaining` and fires `onAllDone` when all requests complete (success or failure). The `cancelAll()` interaction also works correctly -- aborted in-flight requests trigger `onError` through the catch handler, and drained queue entries get `onError` called directly, both paths decrementing the counter.

2. **No mutation of caller objects**: `batch()` now spreads into a new `wrapped` object (`{ ...req, onDone: ..., onError: ... }`) instead of mutating the caller's request objects. Original callbacks are captured in closures.

**Type-check**: `npx tsc --noEmit` passes cleanly.

**Minor observations (non-blocking):**

- `signals.length === 1` check on line 126 is dead code -- `signals` always has at least 2 elements (internalController + rootController). Not a bug, just unnecessary.
- The priority queue uses `Array.sort` on every `pump()` call. Fine for typical batch sizes but could be noted as a future optimization target if queue sizes grow large.
- Errors are silently swallowed when `onError` is not provided (line 193-196). This is an acceptable API design since `onError` is optional, but callers should be aware.

LGTM -- approving.

PR #46, opened 2026-04-19
Add automatic pipeline rebuilding on shader recompilation

### LINKED ISSUE #18: Runtime pipeline rebuilding on shader change

rjwalters:
## Summary

- Adds `shaderPipelineDeps: Map<number, Set<number>>` inside `createGfx` to track which pipeline ids depend on each shader id
- Populates the dep set in `makePipeline` and removes stale entries in `destroyPipeline`
- Adds `rebuildPipelinesForShader(shader)` on the `Gfx` interface that iterates dependents and calls `createRenderPipelineAsync` for each, swapping `slot.gpu` in-place so existing `SgPipeline` handles remain valid
- Falls back to the existing pipeline on `createRenderPipelineAsync` rejection (e.g. bad WGSL), logging a warning and calling the optional `onPipelineRebuildError` callback
- Adds `onPipelineRebuilt` and `onPipelineRebuildError` optional callbacks to the `Gfx` interface (Option A from the curation notes)
- Extracts `buildGpuPipelineDescriptor` helper to avoid duplicating pipeline construction logic between `makePipeline` and `rebuildPipeline`

## Test plan

- [ ] After `makePipeline`, confirm `shaderPipelineDeps` contains the new pipeline id under the shader id
- [ ] After `destroyPipeline`, confirm the pipeline id is absent from the dep set
- [ ] Call `rebuildPipelinesForShader` with valid new shader modules (after `recompileShader` from #16); assert `slot.gpu` is replaced and `onPipelineRebuilt` fires
- [ ] Verify `SgPipeline` handle id is unchanged before and after rebuild
- [ ] Trigger rebuild with broken WGSL; assert `slot.gpu` is unchanged and `onPipelineRebuildError` fires
- [ ] Two pipelines sharing one shader — after recompile assert both are rebuilt
- [ ] While async rebuild is in flight, assert draw calls in interim frames use the old pipeline without errors

Closes #18

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -50,6 +50,8 @@ export function createGfx(
   const samplers = new Map<number, SamplerSlot>();
   const shaders = new Map<number, ShaderSlot>();
   const pipelines = new Map<number, PipelineSlot>();
+  // shader id → set of pipeline ids that reference it
+  const shaderPipelineDeps = new Map<number, Set<number>>();
 
   // Per-frame state
   let encoder: GPUCommandEncoder | null = null;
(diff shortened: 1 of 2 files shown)


rjwalters:
## Loom Judge Review — PR #46: Add automatic pipeline rebuilding on shader recompilation

### Type-check
`tsc --noEmit` passes cleanly for all PR-introduced code. The one existing error (`src/gfx.ts:125` SharedArrayBuffer/ArrayBuffer mismatch) is pre-existing on `main` and not introduced by this PR.

---

### Evaluation

#### Shader-pipeline dependency tracking
`shaderPipelineDeps: Map<number, Set<number>>` is populated correctly in `makePipeline` and cleaned up in `destroyPipeline`. The cleanup path guards with `if (slot)` before accessing `slot.desc.shader.id`, which is correct. The `deps?.delete(pip.id)` optional-chain handles the case where the dep set was never created. Empty dep sets are not pruned from the map after the last pipeline is removed, but that is a minor memory leak and not a correctness issue.

#### `buildGpuPipelineDescriptor` extraction
The helper correctly mirrors the construction logic from `makePipeline`. However, there is a meaningful **divergence bug**: the `buildGpuPipelineDescriptor` helper hard-codes blend factors as string literals (`"src-alpha"`, `"one-minus-src-alpha"`, `"one"`) while `makePipeline` on the actual `pr-46` branch reads them from `c.colorBlend` / `c.alphaBlend` (configurable `BlendFactor` / `BlendOp` enums) and applies `c.writeMask`. The helper thus silently drops per-pipeline blend configuration on rebuild, meaning a rebuilt pipeline will have different blend behavior than the original if non-default blend factors were specified.

Similarly, the helper does not forward `desc.images` bind group layout (group 1: textures + samplers), `desc.msaa`, stencil state, or `desc.depth.stencilFront/Back/ReadMask/WriteMask`. These are all present in the actual `makePipeline` path but absent from `buildGpuPipelineDescriptor`, so any pipeline using those features will be rebuilt incorrectly.

#### Async rebuild + handle stability
The slot-swap pattern (`slot.gpu = newGpuPipeline`) is correct: the `SgPipeline` handle id is unchanged, all existing code that looks up `pipelines.get(pip.id)` keeps working, and in-flight frames using `currentPipeline` (a reference to the old slot object) continue rendering the old GPU pipeline until `commit()` clears it. The snapshot `[...deps]` before the async loop correctly avoids iterator-invalidation issues if destroyPipeline is called concurrently.

#### Error fallback
On `createRenderPipelineAsync` rejection the old `slot.gpu` is preserved (no assignment occurs in the catch branch), which is correct. The `console.warn` and optional `onPipelineRebuildError` callback are both fired. One concern: the error is typed as `unknown`, which is correct TypeScript, but the callback signature `(pip: SgPipeline, error: unknown) => void` passes the raw rejection value without any wrapping — callers will need to type-narrow themselves. This is acceptable but worth documenting.

#### Sequential vs. parallel rebuild
`rebuildPipelinesForShader` rebuilds dependents **sequentially** with `await` in a loop. For a shader shared by N pipelines this means N×(GPU compilation time) latency before the last pipeline swaps. Rebuilding them in parallel with `Promise.all` would be preferable, though sequential is functionally correct and conservative.

#### `onPipelineRebuilt` / `onPipelineRebuildError` callbacks
Declared as mutable properties initialized to `undefined` on the `gfx` object and typed correctly in both `Gfx` interface (`types.ts`) and the implementation. The `gfx.onPipelineRebuilt?.(pipHandle)` optional-chain call is safe.

---

### Issues

**Blocking**
1. `buildGpuPipelineDescriptor` does not reproduce `makePipeline`'s configurable blend factors (`colorBlend`, `alphaBlend`, `writeMask`), image/sampler bind group layout (group 1), MSAA, or stencil state. Any pipeline using these features will be rebuilt with a different descriptor than it was originally created with, silently producing wrong rendering or a GPU validation error.

**Non-blocking (should fix)**
2. Sequential `await` in the rebuild loop should be replaced with `Promise.all([...deps].map(...))` to parallelize GPU compilation.
3. Empty dep sets linger in `shaderPipelineDeps` after the last pipeline referencing a shader is destroyed — add `if (deps.size === 0) shaderPipelineDeps.delete(slot.desc.shader.id)` in `destroyPipeline`.

---

### Verdict
**Changes requested.** The core tracking, handle-stability, and error-fallback logic are sound, but `buildGpuPipelineDescriptor` is a partial copy of `makePipeline` that will silently misrebuild any pipeline with non-default blend, texture bindings, MSAA, or stencil state. This must be fixed before merge.

rjwalters:
## Loom Judge Review -- APPROVED

**Stored-descriptor approach**: Correct. The original `GPURenderPipelineDescriptor` is stored at creation time and shallow-cloned on rebuild with only shader modules swapped. This preserves full fidelity with the original pipeline config (blend, depth/stencil, primitive, cull mode, layout, color targets, label). The core issue from the previous rejection is resolved.

**Type-check**: `tsc --noEmit` passes cleanly.

**Dependency tracking**: `shaderPipelineDeps` is properly populated in `makePipeline` and cleaned in `destroyPipeline` with empty-set pruning to prevent memory leaks.

**Parallel rebuilds**: `Promise.all` with `[...deps]` snapshot is correct -- avoids mutation during async iteration.

**Error handling**: Graceful degradation -- catches `createRenderPipelineAsync` failures, keeps old pipeline, logs warning, fires `onPipelineRebuildError` callback.

**Minor note (non-blocking)**: `destroyShader` doesn't clean `shaderPipelineDeps` for the destroyed shader. This is safe because `rebuildPipeline` early-returns when the shader is not found, but a future cleanup pass could add `shaderPipelineDeps.delete(shd.id)` in `destroyShader` for tidiness.

PR #42, opened 2026-04-19
Wire depth/stencil attachments, store actions, and multi-pass support

rjwalters:
Closes #7

## Summary

- **`StoreAction` enum** (`STORE = 0`, `DISCARD = 1`) added to `types.ts`; `ColorAttachment` gains `storeAction` and `resolveImage`; `PassDesc.depthAttachment` gains `storeAction`; `ImageDesc` gains `sampleCount?: 1 | 4`; `PipelineDesc` gains `multisample?: { count?: 1 | 4 }`
- **`beginPass`** now: guards encoder creation so multiple passes share one encoder per frame; maps `LoadAction.DONTCARE` to `"load"` (omits `clearValue` as required by WebGPU); wires `depthStencilAttachment` from `offscreen.depthImage`; uses `resolveStoreOp` helper for `storeOp`; wires `resolveTarget` on color attachments for MSAA resolve
- **`makeImage`** forwards `sampleCount` to `device.createTexture`; **`makePipeline`** forwards `multisample.count` to `createRenderPipeline`
- `uniformOffset` reset moved from `beginPass` to `commit` to avoid corrupting earlier-pass uniform data in multi-pass frames

## Test plan

- [ ] Depth/stencil pass — render overlapping quads at different Z, confirm occlusion; verify `depthStencilAttachment` is non-null in WebGPU capture
- [ ] `DONTCARE` load action — set `action: LoadAction.DONTCARE`, confirm no validation error and `loadOp === "load"` in capture
- [ ] Store `DISCARD` — transient depth buffer with `storeAction: StoreAction.DISCARD`, confirm no crash or validation error
- [ ] Multiple passes per frame — two `beginPass`/`endPass` pairs before `commit`; confirm one submit with two render passes in WebGPU capture
- [ ] MSAA resolve — 4x MSAA color image + 1x resolve target; confirm resolved image is correct when sampled in a subsequent pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #47, opened 2026-04-19
Add configurable blend, stencil operations, and MSAA support

### LINKED ISSUE #6: Render pipeline creation with full vertex layout, blend, depth/stencil config
## Summary
Create render pipelines with complete vertex layout, color blend state, depth/stencil configuration, and primitive topology.

## Current State
`gfx.ts:makePipeline` builds full `GPURenderPipeline` with vertex buffer layouts, color targets, blend state, depth/stencil, and primitive topology. Uses explicit pipeline layout with a uniform bind group.

## Requirements
- [x] Vertex buffer layout with attributes (`gfx.ts:162-172`)
- [x] Color target and blend state (`gfx.ts:174-180`)
- [x] Depth/stencil configuration (`gfx.ts:219-224`)
- [x] Primitive topology and cull mode (`gfx.ts:213-218`)
- [x] Explicit pipeline layout with uniform bind group (`gfx.ts:183-198`)
- [ ] Configurable blend factors (currently hardcoded to standard alpha blend)
- [ ] Stencil operations
- [ ] Multi-sample anti-aliasing (MSAA)
- [ ] Texture/sampler bind group layout (Group 1)
- [ ] Multiple color targets with independent blend state
- [ ] Pipeline caching/deduplication

## References
- Sokol: `sg_make_pipeline()`
- WebGPU: `GPUDevice.createRenderPipeline()`

rjwalters:
Closes #6

## Summary

- Added `BlendFactor`, `BlendOp`, and `StencilOp` enums to `types.ts`
- Extended `ColorTargetDesc` with `colorBlend`, `alphaBlend` (`BlendComponentDesc`), and `writeMask` for per-target configurable blend
- Added `StencilFaceDesc` interface and extended `DepthStencilDesc` with `stencilFront`, `stencilBack`, `stencilReadMask`, `stencilWriteMask`
- Added `MsaaDesc` interface and `msaa` field on `PipelineDesc`; added `sampleCount` on `ImageDesc`
- Added `images` (slot count) field on `PipelineDesc` for Group 1 texture/sampler bind group layout creation
- Added `resolveImages` on `PassDesc.offscreen` for MSAA resolve targets in `beginPass`
- Added `applyStencilRef(ref: number)` to `Gfx` interface and implementation
- Implemented pipeline deduplication cache (`pipelineCache` Map with stable JSON key)
- Wired Group 1 bind group creation in `applyBindings` when `bind.images` is populated

## Test plan

- [ ] Blend factors: render additive blend (`SRC_COLOR`/`ONE`) and verify brighter output
- [ ] Blend disabled: `blendEnabled: false` produces exact colour, no blend
- [ ] Stencil write + test: first pass writes `0x01` with `passOp: REPLACE`, second pass with `compare: EQUAL` renders only masked region
- [ ] Stencil default: pipeline without stencil fields behaves as before (no regression)
- [ ] MSAA pipeline creation: `makePipeline({ msaa: { count: 4 } })` succeeds, `multisample.count === 4`
- [ ] MSAA render: offscreen MSAA pass with resolve target produces anti-aliased edge
- [ ] Texture bind group: checkerboard image binds and samples correctly in fragment shader
- [ ] Multiple samplers: two images and two samplers bind independently
- [ ] Multiple color targets (MRT): offscreen pass with two `colorImages` and two `colors`
- [ ] Pipeline cache hit: identical `PipelineDesc` returns same `GPURenderPipeline` reference
- [ ] Pipeline cache miss: changed field yields distinct pipeline
- [ ] No regression: existing vertex layout, depth, topology, cull mode tests pass


rjwalters:
## Loom Review — PR #47 (blend, stencil, MSAA)

**Type-check**: The `tsc --noEmit` error (`SharedArrayBuffer` assignability on `writeTexture`) pre-exists on `main` (line 125) and was not introduced by this PR. It is pre-existing technical debt.

---

### Critical Issue: encoder-per-pass regression (blocking)

**File**: `src/gfx.ts`, `beginPass` (line 314)

The base branch guards encoder creation with `if (!encoder)` to support multiple passes per frame sharing one `GPUCommandEncoder`. This PR unconditionally calls `device.createCommandEncoder()` on every `beginPass`, removing that guard. `commit()` still only submits `encoder.finish()` on the final encoder reference. Any frame that issues more than one pass (e.g., a shadow/offscreen prepass followed by the main swapchain pass) will silently drop all but the last pass's work.

```ts
// PR (line 314) — always creates a new encoder, discarding previous one:
beginPass(desc?: PassDesc) {
  encoder = device.createCommandEncoder();  // ← overwrites any previous encoder
  ...
```

This should restore the `if (!encoder)` guard. The `uniformOffset = 0` reset (moved to `beginPass`) should remain in `commit()` for correct per-frame uniform layout, since with a shared encoder, multiple passes share one frame's uniform buffer window.

---

### Significant Issue: `ColorAttachment` breaking API removal

**File**: `src/types.ts`, `ColorAttachment` interface (line 231–234)

The PR silently removes `storeAction` and `resolveImage` from `ColorAttachment`. The base branch has:

```ts
export interface ColorAttachment {
  action?: LoadAction;
  storeAction?: StoreAction;
  color?: [number, number, number, number];
  resolveImage?: SgImage;
}
```

The PR narrows this to just `action` and `color`. Callers using per-attachment `storeAction` or `resolveImage` will silently get `undefined` at runtime. The `storeOp` in the swapchain path (`else` branch) is hardcoded to `"store"` with no way for callers to override it for the swapchain surface. The PR description says "MSAA resolve moved to `offscreen.resolveImages`" — but the swapchain path needs it too. Recommend either keeping `resolveImage` on `ColorAttachment` for the swapchain path or documenting that MSAA on the swapchain surface is unsupported.

---

### Minor Issue: `ImageDesc.sampleCount` type is too loose

**File**: `src/types.ts`, line 135

`MsaaDesc.count` is correctly typed as `1 | 2 | 4 | 8 | 16`. But `ImageDesc.sampleCount` is typed as `number`, allowing invalid WebGPU values at the type level. Recommend `sampleCount?: 1 | 4` (or the same union as `MsaaDesc.count`) for consistency and early validation.

---

### Minor Issue: `pipelineCache` key excludes `label`

**File**: `src/gfx.ts`, `pipelineKey` (line 78–90)

`desc.label` is excluded from the key, so two calls with identical pipeline state but different labels will return the cached pipeline with the first label. This is cosmetic-only since `label` is debug metadata, but worth noting.

---

### Observations (no change required)

- **Blend factors/ops** (`BlendFactor`, `BlendOp`): enum string values match WebGPU spec exactly. Defaults (SRC_ALPHA / ONE_MINUS_SRC_ALPHA / ADD) are correct and backward-compatible.
- **Stencil ops** (`StencilOp`, `StencilFaceDesc`, `buildStencilFace`): all 8 WebGPU stencil ops present, defaults (ALWAYS compare, KEEP ops, 0xFF masks) are correct.
- **MSAA pipeline** (`MsaaDesc`): `multisample.count` and `alphaToCoverageEnabled` wired correctly. `desc.msaa ? { ... } : undefined` correctly omits `multisample` when not requested.
- **Resolve attachment**: `storeOp: resolveImg ? "discard" : "store"` in the offscreen path is correct WebGPU practice for MSAA resolve.
- **Group 1 bind group layout**: interleaved texture (binding `i*2`) + sampler (binding `i*2+1`) pattern is consistent between `makePipeline` and `applyBindings`. Sampler fallback `bind.samplers?.[i] ?? bind.samplers?.[0]` is a nice convenience for single-sampler use.
- **`applyStencilRef`**: correct — `passEncoder.setStencilReference(ref)` is the right WebGPU call and the guard against no active pass is consistent with other apply* methods.

---

### Summary

Two issues require changes before merge:
1. Restore the `if (!encoder)` guard in `beginPass` to prevent silent pass dropping in multi-pass frames.
2. Decide on and document the `ColorAttachment` breaking removal — either restore `storeAction`/`resolveImage` or acknowledge the break in a migration note.

The blend/stencil/MSAA pipeline logic itself is correct and well-structured.

rjwalters:
## Judge Review: Changes Requested

### Critical: Cannot merge -- 10 conflicts with main

This PR branches from `99712ea`, but main has since merged PR #42 ("Wire depth/stencil attachments, store actions, and multi-pass support") which adds overlapping functionality. There are **10 merge conflicts** across both `src/gfx.ts` and `src/types.ts`. The PR must be rebased onto current main before it can be reviewed for merge.

### Issues found in the diff itself

**1. Hardcoded shader entry points (regression)**
The PR uses hardcoded `"vs_main"` / `"fs_main"` in `makePipeline`, but current main already uses `shd.vertexEntry` / `shd.fragmentEntry` (respecting `ShaderDesc.vertexEntry`/`fragmentEntry`). This would be a regression if merged.

**2. Pipeline cache key does not include shader entry points**
`pipelineKey()` includes `desc.shader.id` but not the shader's entry point names. If entry points vary (or if a shader handle is reused after destroy/recreate), the cache could return a wrong pipeline. Entry points should be part of the key, or looked up and included.

**3. `storeAction` uses raw `GPUStoreOp` instead of library enum**
Current main already has a `StoreAction` enum for the public API. The PR uses `GPUStoreOp` directly on `ColorAttachment.storeAction`, coupling the public API to a WebGPU native type. Should use the existing `StoreAction` enum.

**4. Redundant resolve image paths**
The PR adds both `ColorAttachment.resolveImage` (per-attachment) AND `PassDesc.offscreen.resolveImages[]` (array on offscreen) with priority logic between them. This creates two ways to specify the same thing. Current main only has per-attachment `resolveImage`, which is cleaner. Pick one path.

### Recommendation

Rebase onto current main first. Many of the previously-rejected fixes (encoder guard, storeAction, resolveImage) are already on main via PR #42. After rebasing, the PR scope should be reduced to the genuinely new features: configurable blend components, stencil operations, MSAA pipeline support, texture/sampler bind group (Group 1), pipeline cache, `applyStencilRef`, and `writeMask`. Fix the entry point regression and cache key issue during the rebase.

rjwalters (commented):
## Review: PR #47 — Blend, Stencil, and MSAA Support

**Type-check result:** Passes cleanly with `tsc --noEmit` (TypeScript 6.0.3, @webgpu/types 0.1.69).

### Enum & Type Quality

All WebGPU enum mappings verified correct:
- `BlendFactor` values exactly match `GPUBlendFactor` string literals (`"zero"`, `"one"`, `"src"`, `"src-alpha"`, etc.)
- `BlendOp` values exactly match `GPUBlendOperation`
- `StencilOp` values exactly match `GPUStencilOperation`
- Enum naming follows Sokol C conventions (`SG_BLENDFACTOR_SRC_COLOR` -> `SRC_COLOR`, etc.)

The new interfaces (`BlendComponentDesc`, `StencilFaceDesc`, `MsaaDesc`) are well-structured and follow the existing patterns in the codebase. Defaults are sensible (standard alpha blending for color, premultiplied alpha for alpha channel, `KEEP` for stencil ops, `ALWAYS` for stencil compare).

### Issues Requiring Changes

**1. Pipeline cache does not invalidate on `rebuildPipeline` or `destroyPipeline` (correctness bug)**

When `rebuildPipeline` succeeds (e.g., during shader hot-reload), it updates `slot.gpu` to the newly created pipeline, but the `pipelineCache` still holds the *old* `GPURenderPipeline` for that key. If `makePipeline` is called again with the same descriptor, it will return the stale pre-reload pipeline from the cache, defeating hot-reload.

Similarly, `destroyPipeline` removes from the `pipelines` map but not from `pipelineCache`, so destroyed pipelines are never GC'd, and re-creating a pipeline with the same desc will return the destroyed one.

Fix: Either (a) invalidate the cache entry in `rebuildPipeline` and `destroyPipeline`, or (b) remove the cache entirely until proper invalidation is designed. The cache also doesn't account for the fact that each `makePipeline` call creates a *new* `GPUPipelineLayout`, but returns the *cached* pipeline (which was created with a different layout) — `getBindGroupLayout()` on the cached pipeline will return the original creation's layout, not the layout created by the current call.

**2. Breaking rename: `multisample` -> `msaa` on `PipelineDesc`**

This is a property rename that silently breaks any existing consumer code using `multisample`. If this is intentional, that's fine for a pre-1.0 library, but please confirm this is deliberate and that there are no downstream consumers. The `ImageDesc.sampleCount` type also stays limited to `1 | 4` while `MsaaDesc.count` now accepts `1 | 2 | 4 | 8 | 16` — these should be aligned.

**3. `node_modules/.package-lock.json` in the diff**

This file should not be in the PR diff. It appears to be an artifact from the merge commit. The `.gitignore` should include `node_modules/`.

### Minor / Follow-up Items (non-blocking)

**4. Stencil load/store ops in `beginPass` are hardcoded**

Now that stencil is configurable at the pipeline level, the hardcoded `stencilLoadOp: "clear"` / `stencilStoreOp: "store"` in `beginPass` limits multi-pass stencil workflows. Users may want to load from a previous pass (`stencilLoadOp: "load"`). Could be a follow-up PR, but worth noting as a gap.

**5. No `stencilClearValue` exposed**

`stencilClearValue` is not set on the depth/stencil attachment, defaulting to 0 per WebGPU spec. This is fine for most cases but should eventually be configurable alongside `depthAttachment.value`.

**6. Blend defaults are applied even when sub-fields omitted**

If a user sets `blendEnabled: true` without providing `colorBlend`/`alphaBlend`, they get standard alpha blending defaults. This is a good UX choice and matches Sokol's behavior.

### What Looks Good

- Configurable blend per-target with `colorBlend`/`alphaBlend` is clean and ergonomic
- `writeMask` via `GPUColorWrite` flags is a good addition
- `StencilFaceDesc` correctly mirrors `GPUStencilFaceState` with matching defaults
- `applyStencilRef()` is a natural addition to the render API
- `resolveImages` on `PassDesc.offscreen` simplifies MSAA resolve wiring
- Priority fallback (`ca?.resolveImage ?? resolveImages?.[i]`) is well-reasoned

### Verdict

Requesting changes due to issue #1 (pipeline cache invalidation bug). The cache as implemented will cause incorrect behavior during shader hot-reload, which is a core workflow for this library.

rjwalters:
## JUDGE Re-review: APPROVED

All three previously flagged issues have been properly resolved:

### 1. Pipeline cache invalidation -- FIXED
- `pipelineCache.clear()` in `rebuildPipeline()` (gfx.ts line 129) correctly invalidates all cached pipelines when shader modules change during hot-reload.
- `pipelineCache.delete(key)` in `destroyPipeline()` (gfx.ts line 490) removes the specific entry so stale pipelines are not reused.

### 2. `multisample` naming on PipelineDesc -- FIXED
- The field is `multisample?: MsaaDesc` on `PipelineDesc` (types.ts line 238), matching WebGPU naming conventions. The `MsaaDesc` type name is an internal detail; the public API surface is correct.

### 3. node_modules removed from git tracking -- FIXED
- `node_modules/` is in `.gitignore` (line 5).
- `git ls-files node_modules/` returns nothing on the branch.
- The diff only contains deletions of previously-tracked node_modules files.

### Build verification
- `npm ci` succeeds.
- `npx tsc --noEmit` passes with zero type errors.

### Code quality notes
- Pipeline cache key (`pipelineKey()`) covers all behavior-affecting fields; correctly excludes cosmetic `label`.
- `pipelineCache.clear()` on hot-reload is the right approach since shader module identity changes make all shader-keyed entries stale.
- Stencil defaults (`ALWAYS` compare, `KEEP` ops) match WebGPU spec defaults.
- `multisample: undefined` when not specified lets WebGPU use its default `count: 1`; default of 4 when `multisample` is set but `count` is omitted is sensible.
- `resolveImages` fallback logic correctly prioritizes per-attachment over offscreen-level.

**Verdict: Approved** -- ready to merge.

PR #48, opened 2026-04-19
Add Vite HMR integration for live app reloading

### LINKED ISSUE #17: Live app reloading without page refresh
## Summary
Enable hot module replacement (HMR) for the app lifecycle, preserving GPU state across code changes where possible.

## Requirements
- [ ] Vite HMR integration — accept module updates
- [ ] Graceful cleanup of previous instance on HMR
- [ ] Preserve canvas element across reloads
- [ ] Option to preserve GPU device across reloads (avoid re-initialization)
- [ ] State serialization/restoration hooks for app-level state
- [ ] Error overlay for runtime errors during HMR

## Design Notes
The cleanup function returned by `run()` enables teardown. HMR flow: old cleanup → re-import → new `run()`. GPU device re-creation is the bottleneck; keeping the device alive across reloads would significantly speed up iteration.

## References
- Vite: `import.meta.hot.accept()`
- Prior art: React Fast Refresh, webpack HMR

rjwalters:
Closes #17

## Summary

- Adds `src/hmr.ts` with `runWithHMR(desc)` — a drop-in wrapper around `run()` that wires Vite's `import.meta.hot.accept` / `import.meta.hot.dispose` to drive a clean teardown-and-restart cycle on every module update
- Extends `AppDesc` in `src/types.ts` with three optional hooks: `serializeState`, `restoreState`, and `onError`
- Updates `src/app.ts` to call `context.unconfigure()` before `device.destroy()` (deterministic WebGPU state) and wraps the frame loop in a try/catch that stops the RAF loop and routes errors to `onError` (or rethrows)
- Exports `runWithHMR` from `src/index.ts`

## Design notes

- Canvas element is preserved across reloads — no special handling needed since the DOM is not replaced
- State round-trip: `serializeState` is called inside the wrapped `cleanup` callback so it fires just before GPU teardown; `restoreState` is called inside the wrapped `init` callback after the new instance is fully initialized
- Error overlay: uses Vite's `hot.send("vite:error", ...)` when available, with a simple DOM overlay as fallback; the overlay is removed on the next successful `vite:afterUpdate`
- `__remount__` is exported so `hot.accept` in the same module graph can trigger a remount using the current desc without a full page reload

## Test plan

- [ ] Basic HMR cycle: edit `frame()` in a sample app, confirm the canvas updates without a page reload
- [ ] Cleanup correctness: verify the cleanup closure fires exactly once per HMR update (no dangling RAF callbacks)
- [ ] Canvas preservation: verify `canvas.width`/`canvas.height` are unchanged after HMR; `context.configure()` is re-called and rendering resumes
- [ ] State round-trip: implement `serializeState`/`restoreState` tracking a counter; verify counter survives a hot reload
- [ ] Error overlay: introduce a runtime error in `frame()`; verify the overlay appears and the loop stops cleanly
- [ ] Full page reload fallback: with `import.meta.hot` absent, `runWithHMR` falls back to a plain `run()` call with no errors
- [ ] TypeScript: `tsc --noEmit` passes (verified in CI)


rjwalters:
## Loom Judge Review — Vite HMR (#48)

**Verdict: APPROVE**

---

### Type-check
`tsc --noEmit` on the PR branch passes cleanly for all touched files (`hmr.ts`, `app.ts`, `types.ts`, `index.ts`). The single pre-existing error in `gfx.ts` (`ArrayBufferView<ArrayBufferLike>` vs `GPUAllowSharedBufferSource`) is a TypeScript 6 + `@webgpu/types` compatibility issue that predates this PR and is unrelated to the HMR work.

---

### HMR accept/dispose wiring ✅
- `hot.dispose` fires `_cleanup()` (which serializes state via the wrapped cleanup), then nulls `_cleanup` and `_currentDesc`. State is preserved at module scope in `_savedState` and survives into the next module evaluation cycle — this is the correct Vite idiom.
- `hot.accept` checks the incoming module for `__remount__` and calls it, allowing the updated module's closure to drive the remount rather than the stale desc reference. Correct.
- Fallback (no `import.meta.hot`): `runWithHMR` calls `_mount` and returns early, delegating to plain `run()` semantics.

### Cleanup lifecycle ✅
- `running = false` stops the RAF loop immediately on cleanup; `resizeObserver.disconnect()` removes the observer; all event listeners are removed via `eventMap`; `desc.cleanup?.(gfx)` fires; `context.unconfigure()` is called before `device.destroy()` — deterministic WebGPU teardown added in this PR is correct.
- The `_mount` guard (`if (_cleanup) { _cleanup(); }`) prevents double-mount if `_mount` is called while an instance is already live.

### State serialization ✅
- `serializeState` is called inside the wrapped `cleanup` (before GPU teardown), so `gfx` is still valid.
- `restoreState` is called inside the wrapped `init` after `desc.init(gfx)` returns, so GPU resources are available.
- `_savedState` is cleared to `undefined` at the top of `runWithHMR` before use, preventing stale state from leaking across unrelated reloads.

### Error overlay ✅
- Frame-loop errors: `running = false` halts the RAF; `desc.onError` is called if provided, otherwise the error is rethrown (correct—avoids silent swallowing).
- Init errors: caught in `_mount`, routed to `_showErrorOverlay`, `_cleanup` and `_currentDesc` are nulled.
- Overlay: Vite's `hot.send("vite:error", ...)` is tried first; DOM fallback uses a fixed-position overlay with a unique id, deduplicating on re-entry. `hot.on("vite:afterUpdate", ...)` clears it on next successful update.

### Canvas preservation ✅
- Canvas element is not replaced by HMR — `context.unconfigure()` releases the WebGPU swap-chain but the DOM element persists. The next `run()` call re-calls `context.configure()` on the same element. `canvas.width`/`canvas.height` are reset by the `resize()` call in the new instance's init path. Correct.

---

1. `__remount__` is exported from `hmr.ts` but `index.ts` only re-exports `runWithHMR`. Consumers who want to call `__remount__` from a sibling module need to import directly from `sokol-ts/src/hmr.js`. This is intentional per the design notes, but a public re-export or JSDoc note pointing to the import path would reduce confusion.
2. The `hot.on("vite:afterUpdate", ...)` listener registered in `_showErrorOverlay` is never explicitly removed. It fires once and removes the overlay element, but the listener itself stays attached for the session. This is harmless in practice (the callback is a no-op after the element is gone), but a `hot.off` call or a one-shot wrapper would be cleaner.
3. `rebuildPipelinesForShader` and `onPipelineRebuilt`/`onPipelineRebuildError` appear on the `Gfx` interface in `types.ts` (from a previous PR) but are absent from `pr-48:src/types.ts` — this is a branch-base difference, not introduced by this PR.

---

All five evaluated criteria are satisfied. The implementation is sound.

PR #62, opened 2026-04-19
Guard debugText maxChars against index buffer overflow

### LINKED ISSUE #54: Guard debugText maxChars against index buffer overflow
## Problem
debugText uses `Uint16Array` for indices. If `maxChars > 16384`, vertex indices exceed 65535 and silently wrap, producing garbled output. The default (8192) is safe but users can pass larger values.

## Proposed Fix
Clamp `maxChars` to 16384 in the constructor, or switch to `Uint32Array` indices when `maxChars > 16384`.

### The math
- Each character = 4 vertices (quad corners) + 6 indices (two triangles).
- `maxChars` characters = `maxChars * 4` vertices.
- `Uint16Array` max value = 65535, so max vertex index = 65535.
- **Overflow threshold**: `maxChars * 4 > 65535` → `maxChars > 16383` → **`maxChars >= 16384`** causes wrap.

### Note on source file
The `debugText` module is being introduced in PR #50 (not yet merged). The relevant file will likely be `src/debugText.ts` or similar. The fix should be applied in the constructor/setup function where `maxChars` is first consumed.

### Recommended approach (simplest)
In the constructor, clamp early:
```ts
maxChars = Math.min(maxChars, 16383);
```
And optionally `console.warn` if the user-supplied value was larger.

### Alternative approach (more capable)
Dynamically choose index type:
```ts
const IndexArray = maxChars > 16383 ? Uint32Array : Uint16Array;
const indexFormat = maxChars > 16383 ? "uint32" : "uint16";
```
This requires also passing the correct `indexFormat` to `setIndexBuffer` and using `IndexType.UINT32` in the pipeline desc.

## Acceptance Criteria
- [ ] `maxChars > 16383` either clamps with a warning or switches to 32-bit indices
- [ ] No silent wrapping of vertex indices
- [ ] Default `maxChars` (8192) behavior is unchanged

rjwalters:
## Summary

- Clamp `maxChars` to 16383 in `createDebugText()` to prevent `Uint16Array` index buffer overflow
- Emit `console.warn` when the user-supplied value exceeds the limit
- Default `maxChars` (8192) behavior is unchanged

## Details

Each character uses 4 vertices (quad corners). With `Uint16Array` indices, the maximum addressable vertex index is 65535, so the maximum safe `maxChars` is `floor(65535 / 4) = 16383`. Values above this caused vertex indices to silently wrap around, producing garbled rendering.

Closes #54


rjwalters:
## Loom Judge Review: APPROVED

**Correctness:** The fix is mathematically sound. Each character requires 4 vertices (quad corners), and the index buffer uses `Uint16Array` (max value 65535). The maximum safe `maxChars` is `floor(65535 / 4) = 16383`, since character 16383 uses vertex indices 65532–65535 — exactly at the `Uint16` ceiling. Character 16384 would need vertex index 65536, silently wrapping to 0 and producing garbled rendering. The clamping via `Math.min` is the correct approach.

**API consistency:** The default `maxChars` of 8192 is well below the new limit, so existing users are unaffected. The `console.warn` is appropriate — it alerts developers without throwing, which is consistent with the graceful-clamp pattern already used at line 498 (`if (charCount >= maxChars) return`).

**Type safety:** `npx tsc --noEmit` passes cleanly. The `MAX_CHARS_UINT16` constant is correctly typed as `number` by inference.

**Nit (non-blocking):** There is no guard against `maxChars <= 0` or non-integer values. This is a pre-existing concern, not introduced by this PR, so it does not block approval.

No regressions detected. Clean, minimal, well-commented fix.

PR #64, opened 2026-04-19
Fix debugText printf %e/%g format specifiers

### LINKED ISSUE #55: Fix debugText printf %e/%g format specifiers
## Problem
The `%e` and `%g` printf specifiers in debugText fall into the `toFixed()` branch, producing fixed-point output instead of exponential/general notation.

## Proposed Fix
Handle `%e` with `Number.toExponential()` and `%g` with `Number.toPrecision()`.

### Note on source file
The `debugText` module is being introduced in PR #50 (not yet merged). The printf implementation will be in the debugText source file (likely `src/debugText.ts`).

### Expected fix
In the printf format-specifier switch/if-chain, add explicit branches for `%e` and `%g` **before** the `%f` fallthrough:

```ts
case 'e':
case 'E':
  result += Number(arg).toExponential(precision ?? 6);
  break;
case 'g':
case 'G': {
  const p = precision ?? 6;
  const val = Number(arg);
  // %g uses the shorter of %e and %f, stripping trailing zeros
  const fixed = val.toFixed(p);
  const exp = val.toExponential(p - 1);
  result += (exp.length < fixed.length ? exp : fixed).replace(/\.?0+$/, '');
  break;
}
```

Alternatively, `Number.toPrecision(p)` closely matches C `%g` behavior (significant digits, auto-selects notation):
```ts
case 'g':
case 'G':
  result += Number(arg).toPrecision(precision ?? 6);
  break;
```

### JS method reference
| Specifier | C behavior | JS equivalent |
|-----------|-----------|---------------|
| `%f` | Fixed-point | `Number.toFixed(precision)` |
| `%e` | Exponential | `Number.toExponential(precision)` |
| `%g` | Shorter of %e/%f | `Number.toPrecision(precision)` |

## Acceptance Criteria
- [ ] `printf("%e", 123456.789)` outputs `1.234568e+5` (exponential notation)
- [ ] `printf("%g", 0.00001)` outputs `1e-5` (auto-selects exponential)
- [ ] `printf("%g", 1.5)` outputs `1.5` (auto-selects fixed)
- [ ] `printf("%.2e", 3.14159)` respects precision → `3.14e+0`
- [ ] `%f` behavior unchanged

rjwalters:
## Summary
- `%e`/`%E` now correctly use `toExponential()` instead of `toFixed()`
- `%g`/`%G` now correctly use `toPrecision()` instead of `toFixed()`
- Updated JSDoc to list the supported format specifiers

Closes #55


rjwalters:
## Loom Judge Review: Approved

**Correctness**: The fix properly routes `%e`/`%E` to `toExponential()` and `%g`/`%G` to `toPrecision()`, matching the accepted approach from #55. The `precision || 1` guard on line 580 correctly prevents the `RangeError` that `toPrecision(0)` would throw, consistent with C's behavior of treating `%g` precision 0 as 1.

**Type safety**: No issues. Same casting pattern as the existing `%f` branch.

**No regressions**: The `%f`, `%d`/`%i`, `%s`, and `%%` paths are unchanged. The regex character class already included `eEgG` before this PR; only the handler logic changed.

**Type-check**: `npx tsc --noEmit` passes cleanly on a fresh clone.

**Minor note** (non-blocking): `toPrecision()` does not strip trailing zeros the way C's `%g` does (e.g., `(1.0).toPrecision(6)` yields `"1.00000"` whereas C gives `"1"`). This is fine for a debug text utility and is the approach accepted in the issue.

Looks good to merge.

PR #66, opened 2026-04-19
Cache bind groups in applyBindings

### LINKED ISSUE #51: Cache bind groups in applyBindings
## Problem
`applyUniforms` (called after `applyBindings`) allocates a new `GPUBindGroup` via `device.createBindGroup()` on **every call** (line 406 of `src/gfx.ts`). For apps calling this multiple times per frame, this creates unnecessary GPU object churn. Additionally, when texture/sampler bind groups are added to `applyBindings`, the same pattern will apply there.

## Proposed Fix
Cache the bind group keyed on the combination of bound resources. Invalidate the cache entry when any bound resource is destroyed or updated.

### Current allocation site (`src/gfx.ts`)
- **Line 406**: `device.createBindGroup(...)` inside `applyUniforms` — creates a new bind group every frame for the uniform buffer (group 0, binding 0).
- **`applyBindings` (line 378)**: Currently only calls `setVertexBuffer`/`setIndexBuffer` on the pass encoder — no bind groups yet. When texture/sampler bind groups land (group 1+), the same caching strategy should apply.

### Cache key strategy
- For uniform bind groups: key on `(pipelineLayoutId, uniformBufferId, alignedOffset, dataByteLength)`.
- For future texture/sampler bind groups: key on `(pipelineLayoutId, ...imageIds, ...samplerIds)`.
- A simple string key like `"${pipelineId}:${bufId}:${offset}:${size}"` is sufficient; these are hot-path allocations but the key set is small per frame.

### Invalidation approach
- Clear cache entries referencing a resource when `destroyBuffer`, `destroyImage` (line 280), or `destroySampler` (line 283) is called.
- Alternatively, use a `WeakRef`-based cache or a per-frame generation counter — since `uniformOffset` resets each frame (`beginPass`), stale entries naturally expire.

### Relevant types
- `Bindings` interface (`src/types.ts` line 166): `{ vertexBuffers, indexBuffer?, images?, samplers? }`.

## Acceptance Criteria
- [ ] Bind group is reused when the same bindings are applied consecutively
- [ ] Cache is invalidated when bound resources change (destroy/update)
- [ ] No memory leaks from stale cache entries
- [ ] No observable behavior change in existing tests/samples

rjwalters:
## Summary
- Adds two `Map`-based caches for `GPUBindGroup` objects to eliminate redundant `device.createBindGroup()` calls on every frame
- **Uniform bind group cache (group 0)**: keyed on `pipelineId:bindingSize` — the dynamic offset is passed separately to `setBindGroup`, so the bind group is fully reusable across frames
- **Texture/sampler bind group cache (group 1)**: keyed on `pipelineId:imageIds:samplerIds` — reused when the same textures/samplers are bound to the same pipeline

## Cache invalidation
- `destroyImage` / `destroySampler`: scans and removes any cached bind groups referencing the destroyed resource
- `destroyPipeline`: removes all cached bind groups for that pipeline
- `rebuildPipeline` (shader hot-reload): invalidates cached bind groups after the pipeline object is swapped

## Test plan
- [ ] Verify existing samples render identically (triangle, cube, texture demos)
- [ ] Confirm no memory leaks: cache entries are removed on resource destroy
- [ ] Verify hot-reload still works: cache is invalidated when pipelines are rebuilt
- [ ] Check that switching pipelines within a frame uses correct cached bind groups

Closes #51


rjwalters:
## JUDGE Review: Cache bind groups in applyBindings

**Verdict: APPROVED**

### Type-checking

Clean checkout on the branch passes `npx tsc --noEmit` with no errors.

### Correctness

**Uniform bind group cache (group 0):** The cache key `pipelineId:bindingSize` correctly captures the two factors that determine bind group identity. The dynamic offset is passed separately to `setBindGroup`, so it does not affect the bind group object itself. The shared uniform buffer (`ensureUniformBuffer`) is created once and never recreated, so cached bind groups always reference a valid buffer.

**Texture/sampler bind group cache (group 1):** The cache key `pipelineId:imgIds:samplerIds` is correct. Since there are no `updateImage` or `updateSampler` mutation APIs, the GPU resources (texture views, sampler objects) referenced by a cached bind group remain valid for the lifetime of the resource.

### Cache invalidation

All three invalidation paths are correct:

- **`destroyImage` / `destroySampler`:** Scans texture/sampler cache for entries referencing the destroyed resource ID. The segment-based key parsing (`parts[1]` for images, `parts[2]` for samplers) correctly matches the key format.
- **`destroyPipeline`:** Removes all cached entries for that pipeline from both caches using `startsWith(prefix)`. No false positives -- e.g., pipeline ID `1:` won't match `10:` because the colon delimiter prevents it.
- **`rebuildPipeline` (hot-reload):** Invalidates caches after the pipeline GPU object is swapped, so subsequent frames create bind groups against the new pipeline layout. Correctly placed after `slot.gpu = newGpuPipeline` and before the `onPipelineRebuilt` callback.

**Map deletion during iteration:** Deleting from a `Map` while iterating its `.keys()` is safe per the ES6 specification -- the iterator visits all entries that existed at the start and haven't been deleted during iteration. Verified independently.

### Performance

The caching eliminates per-frame `device.createBindGroup()` calls for both uniform and texture/sampler bind groups. Cache lookups are `Map.get()` (O(1)). The cache keys are short strings built from numeric IDs, so construction overhead is negligible compared to the avoided GPU API calls.

The invalidation functions do linear scans over cache entries, but these run only on resource destruction (infrequent), not per-frame. Acceptable tradeoff.

### API consistency

- `currentPipelineId` is correctly maintained in lockstep with `currentPipeline` -- set in `applyPipeline`, reset in `commit`.
- No public API changes -- this is a purely internal optimization.
- The `bindingSize` variable consolidation (`Math.max(data.byteLength, 256)`) used in both the cache key and the offset advance is a nice cleanup.

### No regressions

The behavioral contract is preserved: same bind groups are set on the pass encoder, same dynamic offsets, same buffer writes. The only difference is that bind group objects are reused across frames rather than recreated.

No concerns. Clean approval.

PR #67, opened 2026-04-19
Support separate vertex/fragment shader URLs in fetchShader

### LINKED ISSUE #56: Support separate vertex/fragment shader URLs in fetchShader
## Problem
`ShaderDesc` in `src/types.ts` (line 131-138) supports both `source` (combined VS+FS) and separate `vertexSource`/`fragmentSource` fields. However, there is no fetch helper to load shader files from URLs. Users must manually fetch WGSL files and pass the source strings to `gfx.makeShader()`.

A `fetchShader` utility would reduce boilerplate, and it should support loading separate vertex/fragment shader files in parallel.

## Current State
- `ShaderDesc` already supports `source`, `vertexSource`, and `fragmentSource` (see `src/types.ts:131-138`)
- `gfx.makeShader()` in `src/gfx.ts:165-198` handles both combined and separate modules — it creates one `GPUShaderModule` when `source` is set, or two when `vertexSource`/`fragmentSource` are provided
- No `src/fetch.ts` exists yet; the module needs to be created and exported from `src/index.ts`

## API Proposal

```typescript
// src/fetch.ts

import { type ShaderDesc, type SgShader, type Gfx } from "./types.js";

// Overload 1: single combined shader file
export async function fetchShader(
  gfx: Gfx,
  url: string,
  opts?: { vertexEntry?: string; fragmentEntry?: string; label?: string },
): Promise<SgShader>;

// Overload 2: separate vertex and fragment shader files
export async function fetchShader(
  gfx: Gfx,
  urls: { vertexUrl: string; fragmentUrl: string },
  opts?: { vertexEntry?: string; fragmentEntry?: string; label?: string },
): Promise<SgShader>;

// Implementation: detect object vs string for first arg
export async function fetchShader(
  gfx: Gfx,
  urlOrUrls: string | { vertexUrl: string; fragmentUrl: string },
  opts?: { vertexEntry?: string; fragmentEntry?: string; label?: string },
): Promise<SgShader> {
  if (typeof urlOrUrls === "string") {
    const resp = await fetch(urlOrUrls);
    if (!resp.ok) throw new Error(`Failed to fetch shader: ${resp.statusText}`);
    const source = await resp.text();
    return gfx.makeShader({ source, ...opts });
  } else {
    const [vsResp, fsResp] = await Promise.all([
      fetch(urlOrUrls.vertexUrl),
      fetch(urlOrUrls.fragmentUrl),
    ]);
    if (!vsResp.ok) throw new Error(`Failed to fetch vertex shader: ${vsResp.statusText}`);
    if (!fsResp.ok) throw new Error(`Failed to fetch fragment shader: ${fsResp.statusText}`);
    const [vertexSource, fragmentSource] = await Promise.all([
      vsResp.text(),
      fsResp.text(),
    ]);
    return gfx.makeShader({ vertexSource, fragmentSource, ...opts });
  }
}
```

## Acceptance Criteria
- [ ] Create `src/fetch.ts` with the `fetchShader` function supporting both overloads
- [ ] Export `fetchShader` from `src/index.ts`
- [ ] Single URL path: fetches one file, passes as `source` to `makeShader`
- [ ] Dual URL path: fetches vertex and fragment in parallel via `Promise.all`, passes as `vertexSource`/`fragmentSource`
- [ ] Meaningful error messages on fetch failure (include URL and status)
- [ ] Passes through `vertexEntry`, `fragmentEntry`, and `label` options
- [ ] Type-checks cleanly with `tsc --noEmit`

## Implementation Notes
- Keep the module small and focused; no caching or retry logic in v1
- The `Gfx` interface already has `makeShader(desc: ShaderDesc): Promise<SgShader>` so the fetch helper just wraps it
- Consider also adding `fetchImage` in the same module (future scope, not required for this issue)

rjwalters:
## Summary

- Add `vertexUrl` / `fragmentUrl` fields to `FetchShaderRequest` as a discriminated union alternative to the existing single `url` field
- When both split URLs are provided, fetch vertex and fragment WGSL files in parallel via the existing sfetch queue, then pass them as separate `vertexSource`/`fragmentSource` to `gfx.makeShader()`
- Backward compatible: the existing single-URL form continues to work unchanged

## Changes

- **`src/types.ts`** — Refactored `FetchShaderRequest` into a discriminated union (`FetchShaderSingleRequest | FetchShaderSplitRequest`) with a shared base interface
- **`src/fetch.ts`** — Extended `fetchShader()` to detect split URLs and issue two parallel fetches, combining the results before calling `makeShader()`; also fixed both paths to properly `await` the async `makeShader()` call

## Test plan

- [ ] Verify single-URL `fetchShader` still works (backward compat)
- [ ] Verify `{ vertexUrl, fragmentUrl }` fetches both files and creates a shader with separate modules
- [ ] Verify error in either fetch reports via `onError` and does not call `onDone`
- [ ] `npx tsc --noEmit` passes cleanly

Closes #56

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -236,21 +236,77 @@ export function createSfetch(desc?: FetchSetupDesc): SfetchContext {
     },
 
     fetchShader(gfx: Gfx, req: FetchShaderRequest): void {
-      ctx.fetch<string>({
-        url: req.url,
-        type: "text",
-        signal: req.signal,
-        onProgress: req.onProgress,
-        onError: req.onError,
-        onDone(source, url) {
-          const shader = gfx.makeShader({
-            vertexSource: source,
-            fragmentSource: source,
-            label: req.label,
-          });
-          req.onDone(shader, url);
-        },
-      });
+      if (req.vertexUrl && req.fragmentUrl) {
+        // Separate vertex/fragment shader files — fetch in parallel.
+        let vertexSource: string | undefined;
+        let fragmentSource: string | undefined;
+        let completed = 0;
+        let errored = false;
+
+        const finish = async () => {
+          if (++completed < 2 || errored) return;
+          try {
+            const shader = await gfx.makeShader({
+              vertexSource: vertexSource!,
+              fragmentSource: fragmentSource!,
+              label: req.label,
+            });
+            req.onDone(shader, `${req.vertexUrl}+${req.fragmentUrl}`);
+          } catch (err) {
+            onError(err instanceof Error ? err : new Error(String(err)), `${req.vertexUrl}+${req.fragmentUrl}`);
+          }
+        };
+
+        const onError = (err: Error, url: string) => {
+          if (errored) return;
+          errored = true;
+          if (req.onError) req.onError(err, url);
+        };
+
+        ctx.fetch<string>({
+          url: req.vertexUrl,
+          type: "text",
+          signal: req.signal,
+          onProgress: req.onProgress,
+          onError,
+          onDone(source) {
+            vertexSource = source;
+            finish();
+          },
+        });
+
+        ctx.fetch<string>({
+          url: req.fragmentUrl,
+          type: "text",
+          signal: req.signal,
+          onError,
+          onDone(source) {
+            fragmentSource = source;
+            finish();
+          },
+        });
+      } else {
+        // Single combined WGSL file.
+        ctx.fetch<string>({
+          url: req.url!,
+          type: "text",
+          signal: req.signal,
+          onProgress: req.onProgress,
+          onError: req.onError,
+          async onDone(source, url) {
+            try {
+              const shader = await gfx.makeShader({
+                vertexSource: source,
+                fragmentSource: source,
+                label: req.label,
+              });
+              req.onDone(shader, url);
+            } catch (err) {
+              if (req.onError) req.onError(err instanceof Error ? err : new Error(String(err)), url);
+            }
+          },
+        });
+      }
     },
 
     batch(requests: FetchRequest<unknown>[], onAllDone: () => void): void {
--- a/src/types.ts
+++ b/src/types.ts
@@ -344,14 +344,30 @@ export interface FetchImageRequest {
   label?: string;
 }
 
-export interface FetchShaderRequest {
-  url: string;
+/** Base fields shared by both single-URL and split-URL shader requests. */
+interface FetchShaderRequestBase {
   signal?: AbortSignal;
   onProgress?: (p: FetchProgress) => void;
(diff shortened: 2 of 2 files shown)


rjwalters:
## JUDGE Review: Approved

**Type-check:** Passes cleanly (`npx tsc --noEmit` with `@webgpu/types@0.1.68` -- the 0.1.69 dist issue is pre-existing, not introduced by this PR).

### Types (`src/types.ts`)

The discriminated union approach is well-designed:

- `FetchShaderSingleRequest` requires `url: string` and disallows `vertexUrl`/`fragmentUrl` via `?: undefined`
- `FetchShaderSplitRequest` requires both `vertexUrl: string` and `fragmentUrl: string`, disallowing `url`
- The `?: undefined` sentinel pattern is the standard TypeScript idiom for making discriminated unions work with property checks, and it works correctly here

The base interface is kept private (not exported), which is the right call -- consumers only need the concrete variants and the union alias.

### Implementation (`src/fetch.ts`)

**Correctness:**
- The split-URL path correctly uses a `completed` counter and `errored` flag to coordinate two parallel fetches. The `finish()` function gates on `completed === 2` and `!errored` before calling `makeShader()`.
- The `errored` guard prevents double-reporting errors (once per fetch) and prevents `finish()` from proceeding after a failure.
- Both paths now properly `await` the async `makeShader()` call and handle errors with try/catch -- this is a real bug fix over the original code which was fire-and-forget.

**One minor observation (non-blocking):** The `onProgress` callback is only wired to the vertex fetch in the split path (line 270), not the fragment fetch (line 279-286). This is a reasonable simplification since progress tracking across two parallel downloads is inherently ambiguous, but it could be worth a brief comment noting the intentional asymmetry.

**URL concatenation for `onDone`:** Using `${req.vertexUrl}+${req.fragmentUrl}` as the URL string passed to `onDone` is a pragmatic choice. The `+` separator is unlikely to appear in real URLs, making it a reasonable convention.

### Backward compatibility

The single-URL code path is preserved with only the addition of proper `await` and error handling for `makeShader()`. The `req.url!` non-null assertion on line 291 is safe because TypeScript narrows the type: if we're in the `else` branch, `req.vertexUrl` and `req.fragmentUrl` are both falsy, so the union narrows to `FetchShaderSingleRequest` where `url` is required.

### API consistency with Sokol conventions

The naming (`vertexSource`/`fragmentSource` in `ShaderDesc`, `vertexUrl`/`fragmentUrl` in `FetchShaderRequest`) follows the existing camelCase pattern and the vertex/fragment terminology already established in the codebase.

**Verdict: Approve.** Clean, well-structured change with good type safety. The `makeShader` await fix is a valuable bonus.

PR #85, opened 2026-04-21
fix: correct uniform staging buffer limits and bind group size

### LINKED ISSUE #82: Fix limits on staging buffer
Confirmed: this is a sokol-ts-specific port gap, not a fundamental limit from sokol or WebGPU. Upstream sokol-C defaults to 4 MB (64× bigger) and is user-configurable via sg_desc.uniform_buffer_size. Here's the exact issue text:

Issue 8: Uniform staging buffer is 64 KB / hardcoded — 64× smaller than upstream sokol default
Where it happens

[src/gfx.ts:388](https://github.com/rjwalters/sokol-ts/blob/main/src/gfx.ts#L388):

const UNIFORM_BUFFER_SIZE = 65536; // 64KB uniform staging per ring slot
With the 256-byte dynamic-offset alignment enforced by WebGPU, this caps us at 256 applyUniforms() calls per pass. Exceeding it throws:

Uniform buffer overflow: offset 65536 + 256 exceeds UNIFORM_BUFFER_SIZE (65536).
Too many applyUniforms calls in a single pass.
Repro: [the repo linked in comments]. A saturation benchmark with 8192 distinct per-object uniforms needed 32 passes (256 draws each) just to fit.

Comparison with upstream sokol-gfx-C

Sokol-gfx-C's WebGPU backend docs:

The sokol-gfx WebGPU backend uses the sg_desc.uniform_buffer_size item to allocate a single per-frame uniform buffer which must be big enough to hold all data written by sg_apply_uniforms() during a single frame, including a worst-case 256-byte alignment. The default size is 4 MB, which is enough for 16384 sg_apply_uniform() calls per frame.

So upstream:

Default: 4 MB → ~16,384 calls per frame (not per pass).
User-configurable via sg_desc.uniform_buffer_size.
Scoped per frame, not per pass.
sokol-ts v0.1:

Hardcoded 64 KB → 256 calls per pass.
Not configurable.
Scoped per pass (different from upstream).
Why the current check is overly restrictive

[gfx.ts:389](https://github.com/rjwalters/sokol-ts/blob/main/src/gfx.ts#L389) asserts:

if (UNIFORM_BUFFER_SIZE > device.limits.maxUniformBufferBindingSize) { … }
This conflates two separate WebGPU limits. The staging GPUBuffer can be arbitrarily large; only the bound window is limited by maxUniformBufferBindingSize (spec-guaranteed ≥ 64 KB). Since the backend already uses dynamic offsets (setBindGroup(…, dynamicOffsets)), each bind is only 256 bytes — so a 4 MB staging buffer with 256-byte windows is perfectly legal and matches sokol-C's behavior.

Proposed fix

Raise the default to 4 MB to match upstream.
Drop the > maxUniformBufferBindingSize check — it's comparing the wrong limit.
Expose it via RunOpts/GfxDesc:
export interface RunOpts {
  canvas: string | HTMLCanvasElement;
  init: (gfx: Gfx) => Promise<void> | void;
  frame: (gfx: Gfx) => void;
  uniformBufferSize?: number; // bytes; default 4 * 1024 * 1024
}
Align scope with upstream: track usage per frame (reset at commit()), not per pass. This avoids surprising the user who splits work across many passes and finds a 256-call limit each.
Impact

With the current 256-call cap, any renderer that issues more than ~256 draw calls with distinct per-object uniforms has to split into multiple passes, paying extra beginPass/endPass overhead. Real scenes hit this very easily (e.g. a town with 300+ dynamic objects). Upstream sokol users wouldn't expect to hit it at all.

Tagging suggestion: enhancement, backend:webgpu, parity:upstream.

rjwalters:
## Summary
Fixes multiple issues with the uniform staging buffer: undersized default, incorrect bind group size field, per-pass scope instead of per-frame, and broken validation. Also adds a configurable `uniformBufferSize` option to `AppDesc`.

## Changes
- Increase default uniform buffer from 64 KB to 4 MB (matching upstream sokol-C)
- Fix bind group `size` field from full buffer size to per-slot size (256 bytes) -- this was the critical bug that would cause WebGPU validation errors with larger buffers
- Change `uniformOffset` reset from `beginPass` (per-pass) to `commit` only (per-frame), matching upstream sokol-C semantics
- Replace incorrect `maxUniformBufferBindingSize` validation (was comparing total buffer size; now validates per-slot binding size)
- Add `uniformBufferSize` option to `AppDesc` and thread it through `app.ts` to `createGfx`
- Update overflow error message to say "per frame" instead of "per pass"

| Default uniform buffer size is 4 MB | Done | `DEFAULT_UNIFORM_BUFFER_SIZE = 4 * 1024 * 1024` |
| Validation checks per-slot binding size, not total buffer | Done | Guard now compares `UNIFORM_SLOT_SIZE` (256) against `maxUniformBufferBindingSize` |
| Bind group `size` field uses per-slot size (256) | Done | Changed from `UNIFORM_BUFFER_SIZE` to `UNIFORM_SLOT_SIZE` at bind group creation |
| `uniformOffset` reset only in `commit()` | Done | Removed reset from `beginPass`; kept in `commit` |
| `AppDesc` gains `uniformBufferSize` field | Done | Added with JSDoc in `types.ts` |
| `createGfx` accepts and uses configurable size | Done | New `options` parameter with fallback to 4 MB default |
| Overflow error says "per frame" | Done | Updated string literal |
| No regressions | Done | `pnpm check:ci` passes, all 93 tests pass |

## Test Plan
- `pnpm check:ci` passes (tsc, lint, build)
- All 93 existing unit tests pass
- Type check clean with `tsc --noEmit`

Closes #82

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -106,7 +106,9 @@ export async function run(desc: AppDesc): Promise<() => void> {
 
   resize();
 
-  const gfx = createGfx(device, canvas, context, format);
+  const gfx = createGfx(device, canvas, context, format, {
+    uniformBufferSize: desc.uniformBufferSize,
+  });
 
   // Hoist running flag so the device.lost handler can reference it
   let running = true;
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -352,6 +352,7 @@ interface PipelineSlot {
  * @param canvas - The `HTMLCanvasElement` to render into.
  * @param context - The `GPUCanvasContext` obtained from the canvas.
  * @param format - The preferred swapchain texture format (e.g. from `navigator.gpu.getPreferredCanvasFormat()`).
+ * @param options - Optional configuration. `uniformBufferSize` sets the per-frame uniform staging buffer size (default 4 MB).
  * @returns A fully initialised {@link Gfx} instance.
  */
 export function createGfx(
@@ -359,6 +360,7 @@ export function createGfx(
   canvas: HTMLCanvasElement,
   context: GPUCanvasContext,
   format: GPUTextureFormat,
+  options?: { uniformBufferSize?: number },
 ): Gfx {
   const bufferPool   = new Pool<BufferSlot>(128);
   const imagePool    = new Pool<ImageSlot>(128);
@@ -385,10 +387,14 @@ export function createGfx(
   // Texture/sampler bind group cache (group 1): keyed on "pipelineId:img1,img2,...:smp1,smp2,..."
   const textureSamplerBindGroupCache = new Map<string, GPUBindGroup>();
 
-  const UNIFORM_BUFFER_SIZE = 65536; // 64KB uniform staging per ring slot
-  if (UNIFORM_BUFFER_SIZE > device.limits.maxUniformBufferBindingSize) {
+  const DEFAULT_UNIFORM_BUFFER_SIZE = 4 * 1024 * 1024; // 4 MB, matching upstream sokol-C
+  const UNIFORM_BUFFER_SIZE = options?.uniformBufferSize ?? DEFAULT_UNIFORM_BUFFER_SIZE;
+  // The per-slot binding size (256 bytes) is the window exposed to the shader
+  // via dynamic offsets. Validate that this fits within the device limit.
+  const UNIFORM_SLOT_SIZE = 256;
+  if (UNIFORM_SLOT_SIZE > device.limits.maxUniformBufferBindingSize) {
     throw new Error(
-      `UNIFORM_BUFFER_SIZE (${UNIFORM_BUFFER_SIZE}) exceeds device limit maxUniformBufferBindingSize (${device.limits.maxUniformBufferBindingSize})`
+      `Uniform slot size (${UNIFORM_SLOT_SIZE}) exceeds device limit maxUniformBufferBindingSize (${device.limits.maxUniformBufferBindingSize})`
     );
   }
   let uniformFrameIndex = 0;
@@ -414,7 +420,7 @@ export function createGfx(
     uniformBuffers.push(buf);
     uniformBindGroups.push(device.createBindGroup({
       layout: uniformBindGroupLayout,
-      entries: [{ binding: 0, resource: { buffer: buf, size: UNIFORM_BUFFER_SIZE } }],
+      entries: [{ binding: 0, resource: { buffer: buf, size: UNIFORM_SLOT_SIZE } }],
       label: `uniform_bind_group_${i}`,
     }));
   }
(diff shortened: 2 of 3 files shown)


rjwalters:
**Approved**

This PR correctly addresses all six items raised in issue #82. The changes are minimal, well-targeted, and match upstream sokol-C semantics.

## Code Review

**Bind group `size` field fix (the critical bug):** The `size` field in `createBindGroup` entries specifies the *binding window* visible to the shader, not the total buffer size. Changing this from `UNIFORM_BUFFER_SIZE` (previously 64KB, now 4MB) to `UNIFORM_SLOT_SIZE` (256 bytes) is correct. With `hasDynamicOffset: true`, WebGPU validates that `dynamicOffset + size <= buffer.size`, and the shader only ever sees a 256-byte window. Using the full buffer size here would have caused a WebGPU validation error (`size` would exceed `maxUniformBufferBindingSize` on most devices). This was the most important fix.

**Validation fix:** The old guard compared `UNIFORM_BUFFER_SIZE` against `maxUniformBufferBindingSize`, which is wrong -- that limit applies to the per-binding window, not the total buffer. The new guard correctly validates `UNIFORM_SLOT_SIZE` (256) against that limit. Since the WebGPU spec guarantees `maxUniformBufferBindingSize >= 65536`, this check will never fire on conformant implementations, but it is the correct defensive check.

**Per-frame scope:** Removing `uniformOffset = 0` from `beginPass` and keeping it only in `commit()` correctly matches upstream sokol-C, which resets the staging offset once per frame. This lets users issue `applyUniforms` across multiple passes within a single frame without hitting an artificial per-pass cap.

**Default size increase:** 64KB to 4MB matches sokol-C's default. With 256-byte alignment this allows ~16,384 `applyUniforms` calls per frame, matching the upstream documentation.

**Configurable `uniformBufferSize`:** Clean threading from `AppDesc` through `app.ts` to `createGfx`. The JSDoc is accurate and helpful.

**Error message update:** "per frame" instead of "per pass" -- consistent with the scope change.

## Minor Observation (non-blocking)

The JSDoc on `uniformBufferSize` states "Must be a multiple of 256" but this is not validated at runtime. A non-aligned value would still work (the buffer just would not use the tail bytes past the last 256-byte-aligned slot), so this is purely a documentation note, not a correctness issue. If you want strict enforcement later, a simple guard could be added, but it is fine as-is.

## Test Execution

**Test plan from PR description:**
1. `pnpm check:ci` -- Executed: tsc and build pass cleanly
2. All 93 existing unit tests pass -- Executed: `pnpm test` runs 5 test files, 93 tests, all pass
3. Type check clean with `tsc --noEmit` -- Executed: passes (included in `check:ci`)

## CI Verification

- GitHub Actions: Build and unit tests (Node 20) pass, Build and unit tests (Node 22) pass, E2E tests pass
- Merge state: CLEAN

PR #98, opened 2026-04-21
fix: move _frameStats reset from beginPass to commit for multi-pass accumulation

### LINKED ISSUE #92: frameStats resets on every beginPass, losing multi-pass and compute stats
## Problem

`beginPass()` in `src/gfx.ts` resets `_frameStats` on every call. This means:
- In multi-pass rendering (e.g., the benchmark in #79), only the last pass's stats survive
- Compute-then-render workflows zero out `dispatchCalls` when the render pass begins

Discovered during PR #89 (benchmark) and PR #91 (compute shaders) reviews.

## Expected behavior

`frameStats` should accumulate across all passes in a frame and only reset in `commit()`.

## Affected files

- `src/gfx.ts` — `beginPass()` stats reset logic

rjwalters:
## Summary
Moves the `_frameStats` reset from `beginPass()` to `commit()` so that draw and dispatch statistics accumulate across all render and compute passes within a single frame.

## Changes
- Removed `_frameStats` reset from `beginPass()` in `src/gfx.ts` (was line 1333)
- Added `_frameStats` reset to `commit()` in `src/gfx.ts` (after `_frameCount++`)
- Updated existing test comment and "starts dispatchCalls at zero" test to use `commit()` for reset
- Added test: compute + render pass accumulation (dispatch stats survive `beginPass()`)
- Added test: `commit()` resets all stat fields to zero

| _frameStats accumulates across all beginPass and beginComputePass calls | Pass | New accumulation test verifies dispatchCalls survive a subsequent beginPass |
| _frameStats resets to zero in commit() | Pass | New reset test checks all four fields are zero after commit |
| frameStats returns accumulated totals between endPass and commit | Pass | Accumulation test reads stats after endPass, before commit |
| Multi-pass scenario: two passes report summed stats | Pass | Compute-then-render test confirms 2 dispatchCalls across passes |
| Compute-then-render: dispatch calls survive subsequent beginPass | Pass | Explicit test for this scenario |
| No regression in single-pass rendering | Pass | All 142 existing tests pass |

## Test Plan
- `pnpm check:ci` passes (lint + build)
- All 142 unit tests pass across 6 test files
- 3 new/updated tests in `tests/compute.test.ts` specifically cover multi-pass accumulation and commit-reset behavior

Closes #92

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -1330,7 +1330,6 @@ export function createGfx(
       passEncoder = encoder.beginRenderPass(passDescGpu);
       boundVertexBuffers = [];
       boundIndexBuffer = null;
-      _frameStats = { drawCalls: 0, totalElements: 0, indirectDrawCalls: 0, dispatchCalls: 0 };
     },
 
     applyPipeline(pip: SgPipeline) {
@@ -1587,6 +1586,7 @@ export function createGfx(
       frameTime = (now - lastFrameTime) / 1000;
       lastFrameTime = now;
       _frameCount++;
+      _frameStats = { drawCalls: 0, totalElements: 0, indirectDrawCalls: 0, dispatchCalls: 0 };
     },
 
     async rebuildPipelinesForShader(shader: SgShader): Promise<void> {
--- a/tests/compute.test.ts
+++ b/tests/compute.test.ts
@@ -296,18 +296,63 @@ describe("Compute: DrawStats.dispatchCalls", () => {
     gfx.dispatchWorkgroups(1, 1, 1);
     gfx.endPass();
 
-    // frameStats snapshot should reflect 3 dispatches
-    // Note: beginPass resets stats, but beginComputePass does not reset them
-    // directly. The stats are reset at the first beginPass or stay accumulated.
+    // frameStats accumulates across all passes within a frame;
+    // stats are reset in commit(), not beginPass().
     const stats = gfx.frameStats;
     expect(stats.dispatchCalls).toBe(3);
   });
 
-  it("starts dispatchCalls at zero", () => {
+  it("starts dispatchCalls at zero after commit", () => {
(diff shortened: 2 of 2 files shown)


rjwalters:
## Approved

**Code change review**: The fix is correct and minimal. Moving the `_frameStats` reset from `beginPass()` (line 1333 in the original) to `commit()` (after `_frameCount++`) ensures that draw/dispatch statistics accumulate across all render and compute passes within a single frame, then reset at frame boundaries. This matches the expected frame lifecycle: begin passes, accumulate stats, commit (reset for next frame).

Key observations:

- `beginComputePass()` never had a stats reset, so before this fix, a compute-then-render sequence would lose the compute stats when `beginPass()` was called for the render pass. The fix eliminates this asymmetry.
- The reset location in `commit()` is correct -- it sits after `_frameCount++` and frame timing updates, so `frameStats` remains readable between the last `endPass()` and `commit()`.
- All four stat fields (`drawCalls`, `totalElements`, `indirectDrawCalls`, `dispatchCalls`) are reset together, which is consistent with the initial declaration at line 423.

**Test review**: The 4 test changes are well-structured:

1. Updated accumulation test comment to reflect the new reset semantics
2. "starts dispatchCalls at zero after commit" -- correctly uses `commit()` instead of `beginPass()` as the reset trigger
3. "accumulates dispatch and draw stats across compute and render passes" -- validates the core fix by dispatching in a compute pass, then calling `beginPass()` for a render pass, and confirming dispatch stats survive
4. "resets all stats to zero after commit" -- comprehensive check of all four stat fields after `commit()`

**CI status**: All checks pass (Node 20, Node 22, E2E). Merge state is CLEAN.

**Test scoping**: Full suite (vitest run) -- 142 tests, all passed. Only 2 source files changed, no config file modifications.

PR #101, opened 2026-04-21
feat: cache compute bind groups to avoid redundant GPU allocations

### LINKED ISSUE #94: applyComputeBindings creates a new bind group on every call
## Problem

`applyComputeBindings()` in `src/gfx.ts` creates a new `GPUBindGroup` on every call. For hot compute dispatch loops, this is a performance concern — bind groups could be cached and reused when the bindings haven't changed, similar to how render bind groups are handled.

Discovered during PR #91 (compute shaders) judge review.

## Expected behavior

Cache compute bind groups and only recreate when bindings change, matching the render path's caching strategy.

## Affected files

- `src/gfx.ts` — `applyComputeBindings()`

rjwalters:
## Summary

`applyComputeBindings()` was calling `device.createBindGroup()` unconditionally on every invocation, even when the same pipeline and storage buffers were bound. This adds a `computeBindGroupCache` that follows the existing `textureSamplerBindGroupCache` pattern from the render path.

## Changes

- Added `computeBindGroupCache` (`Map<string, GPUBindGroup>`) keyed by `"{computePipelineId}:{sbufId1},{sbufId2},..."`
- Added `currentComputePipelineId` state variable, set in `applyComputePipeline`, reset in `endPass`
- Modified `applyComputeBindings` to check cache before calling `device.createBindGroup()`
- Added `invalidateComputeCacheForBuffer()` helper, called from `destroyBuffer`
- Added `invalidateComputeCacheForPipeline()` helper, called from `destroyComputePipeline`
- Both caches cleared in `shutdown()`
- Added 4 unit tests: cache hit, cache miss on different bindings, invalidation on buffer destroy, invalidation on pipeline destroy

| Returns cached GPUBindGroup for same pipeline+buffers | Pass | Unit test: spy confirms createBindGroup called once for two identical applyComputeBindings calls |
| createBindGroup called only once for repeated bindings | Pass | Unit test with vi.spyOn call count assertion |
| Cache invalidated on storage buffer destroy | Pass | Unit test: destroyBuffer triggers new createBindGroup on rebind |
| Cache invalidated on compute pipeline destroy | Pass | Unit test: destroyComputePipeline triggers new createBindGroup on rebind |
| Cache cleared on shutdown | Pass | textureSamplerBindGroupCache.clear() and computeBindGroupCache.clear() called at top of shutdown() |
| Existing compute tests pass | Pass | All 144 tests pass |
| No regression in render path caching | Pass | All existing tests pass unchanged |

## Test Plan

- `npx vitest run` -- all 144 tests pass (6 test files)
- TypeScript compiles cleanly (`npx tsc --noEmit`)

Closes #94

### DIFF
--- a/src/gfx.ts
+++ b/src/gfx.ts
@@ -412,6 +412,7 @@ export function createGfx(
   let passEncoder: GPURenderPassEncoder | null = null;
   let computePassEncoder: GPUComputePassEncoder | null = null;
   let currentComputePipeline: ComputePipelineSlot | null = null;
+  let currentComputePipelineId = 0;
   let currentPipeline: PipelineSlot | null = null;
   let currentPipelineId = 0;
   let frameTime = 0;
@@ -424,6 +425,9 @@ export function createGfx(
 
   // Texture/sampler bind group cache (group 1): keyed on "pipelineId:img1,img2,...:smp1,smp2,..."
   const textureSamplerBindGroupCache = new Map<string, GPUBindGroup>();
+
+  // Compute bind group cache: keyed on "computePipelineId:sbuf1,sbuf2,..."
+  const computeBindGroupCache = new Map<string, GPUBindGroup>();
 
   const DEFAULT_UNIFORM_BUFFER_SIZE = 4 * 1024 * 1024; // 4 MB, matching upstream sokol-C
   const UNIFORM_BUFFER_SIZE = options?.uniformBufferSize ?? DEFAULT_UNIFORM_BUFFER_SIZE;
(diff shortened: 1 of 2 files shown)


rjwalters:
## Code Review: Approved

**All CI checks pass. Merge state is CLEAN. 144/144 tests pass locally.**

### Correctness

The implementation is correct and follows the existing `textureSamplerBindGroupCache` pattern faithfully:

- **Cache key construction** (`"{computePipelineId}:{sbufId1},{sbufId2},..."`) is sound. The colon separator between pipeline ID and buffer IDs prevents ambiguity, and comma-separated buffer IDs correctly capture binding order (since `map(h => h.id).join(",")` preserves the array order from `storageHandles`). Two different binding configurations will always produce different keys.

- **`currentComputePipelineId` lifecycle** is correct: set in `applyComputePipeline`, reset to 0 in `endPass`, ensuring stale pipeline IDs cannot leak across passes.

- **Cache invalidation is complete**:
  - `destroyBuffer` calls both `invalidateTextureSamplerCacheForResource(buf.id, "sbuf")` (for render-path storage buffers) and `invalidateComputeCacheForBuffer(buf.id)` (for compute-path). Both caches are covered.
  - `destroyComputePipeline` calls `invalidateComputeCacheForPipeline(pip.id)` using the `startsWith(prefix)` pattern, matching the render-path `invalidateCachesForPipeline`.
  - `shutdown()` clears both caches at the top, before destroying GPU resources.

- **Iteration-while-deleting safety**: Both `invalidateComputeCacheForBuffer` and `invalidateComputeCacheForPipeline` iterate `computeBindGroupCache.keys()` and call `.delete()` during iteration. This is safe in JavaScript -- `Map.keys()` returns an iterator that handles concurrent deletions correctly per the ES6 spec.

### Design

The approach of a separate `computeBindGroupCache` (rather than extending the existing `textureSamplerBindGroupCache` with more complex keys) is the right call. Compute bind groups use a different layout (`storageBindGroupLayout` vs `getBindGroupLayout(1)`) and different key structure, so a separate cache keeps things clean and avoids overloading the render-path cache key format.

### Tests

The 4 new tests are well-structured:
- Cache hit verification via `vi.spyOn` call count comparison (not absolute counts, which avoids brittleness from pipeline-setup bind groups)
- Cache miss on different bindings
- Buffer destroy invalidation
- Pipeline destroy invalidation

All use relative call-count assertions (`callsAfterFirst` pattern), which is robust against internal implementation details that may create bind groups during setup.

### Minor Observations (non-blocking)

No issues found. The implementation is clean and the PR description is thorough with a complete acceptance criteria table.

## Test Execution

1. `npx vitest run` -- Executed: All 144 tests pass (6 test files, 665ms)
2. TypeScript compilation (`npx tsc --noEmit`) -- Verified via CI: passes

## Test Scoping

**Strategy**: Full suite (only 2 files changed, fast test suite)
**Result**: 144 tests, all passed

